PR #16, opened 2017-03-07
Support fixed-length strings in collection_view

kallanreed:
determine the actual string length like we're doing for std::string and std::wstring parsing

Signed-off-by: Kyle Reed <kallanreed@outlook.com>


kallanreed:
Fixes #17

zacbrown:
My only thoughts are, in general, whether it is a good idea to use basic_string in this context given its usage in the STL for std::basic_string.

PR #22, opened 2017-04-25
Ensure that comparers.hpp is compiled as unmanaged code.

zacbrown:
Ensure that comparers.hpp is compiled as unmanaged when /clr is set. This is necessary to ensure that when boost::* makes use of the std::locale lock, we only take the lock on the native side. If the std::locale lock is taken on the managed side, we introduce a loader lock.

Signed-off-by: Zac Brown (ODSP SECURITY) <zbrown@microsoft.com>


zacbrown:
side note: This introduces a thunk for every call to functions in this file. Investigating the overall cost.

PR #41, opened 2018-03-15
Improve property parsing

aydany:
This set introduces two improvements:

1. The properties cache in the parser could store multiple copies of the same property
This situation could occur because the last property parsed was added to the cache and returned but the buffer pointer and property index were not incremented. This is now fixed: every parsed property can exist in the cache just once and there is no duplicate property parsing.

2. String values (ANSI and UNICODE) that belong to the last property in an event record are not guaranteed to be null-terminated. The code assumes that every string is null-terminated, which causes strlen and wcslen to potentially access invalid memory looking for a null terminator. If the memory access succeeds, the parser will throw an exception as the length of the string property will be too large. These error conditions were exposed when processing events from http.sys. I modeled the fix after the way Microsoft.Diagnostics.Tracing.TraceEventRawReaders deals with strings. All unit tests still pass and the events from http.sys are now processed without errors.

### DIFF
--- a/krabs/krabs/parser.hpp
+++ b/krabs/krabs/parser.hpp
@@ -179,13 +179,15 @@ namespace krabs {
             property_info propInfo(pBufferIndex_, currentPropInfo, propertyLength);
             cache_property(pName, propInfo);
 
+            // advance the buffer index since we've already processed this property
+            pBufferIndex_ += propertyLength;
+
             // The property was found, return it
             if (name == pName) {
+                // advance the index since we've already processed this property
+                ++i;
                 return propInfo;
             }
-
-            // Not found yet, advance the buffer index and try again
-            pBufferIndex_ += propertyLength;
         }
 
         // property wasn't found, return an empty propInfo
--- a/krabs/krabs/size_provider.hpp
+++ b/krabs/krabs/size_provider.hpp
@@ -117,8 +117,8 @@ namespace krabs {
         {
             if (propertyInfo.nonStructType.InType == TDH_INTYPE_UNICODESTRING)
             {
-                const wchar_t* p = (const wchar_t*)propertyStart;
-                const wchar_t* pEnd = (const wchar_t*)pRecordEnd;
+                auto p = (const wchar_t*)propertyStart;
+                auto pEnd = (const wchar_t*)pRecordEnd;
                 while (p < pEnd) {
                     if (!*p++) {
                         break;
@@ -128,8 +128,8 @@ namespace krabs {
             }
             else if (propertyInfo.nonStructType.InType == TDH_INTYPE_ANSISTRING)
             {
-                const char* p = (const char*)propertyStart;
-                const char* pEnd = (const char*)pRecordEnd;
+                auto p = (const char*)propertyStart;
+                auto pEnd = (const char*)pRecordEnd;
                 while (p < pEnd) {
                     if (!*p++) {
                         break;
--- a/krabs/krabs/size_provider.hpp
+++ b/krabs/krabs/size_provider.hpp
@@ -39,7 +39,8 @@ namespace krabs {
     private:
         static ULONG get_heuristic_size(
             const BYTE*,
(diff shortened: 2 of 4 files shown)


swannman:
@aydany thanks for the pull request! I'm reviewing now. Also cc @zacbrown 

Here's the TraceEventRawReaders reference: https://github.com/Microsoft/perfview/blob/master/src/TraceEvent/TraceEvent.cs#L3886

swannman:
@aydany would you be willing to write a unit test that covers the non-null-terminated case so we have test coverage on this new code?

krabs/krabs/parser.hpp
```diff
@@ -179,13 +179,15 @@ namespace krabs {
             property_info propInfo(pBufferIndex_, currentPropInfo, propertyLength);
             cache_property(pName, propInfo);
 
+            // advance the buffer index since we've already processed this property
+            pBufferIndex_ += propertyLength;
+
             // The property was found, return it
             if (name == pName) {
+                // advance the index since we've already processed this property
+                ++i;
```
swannman: is this line necessary? if we're about to `return propInfo` then `i` will go out of scope so we don't need to advance it?

aydany: Note that `i` is defined as a reference to  lastPropertyIndex_` at the start of the for loop.

aydany:
@swannman I'll take a look at adding a test tomorrow. Thanks for the review!

krabs/krabs/size_provider.hpp
```diff
@@ -105,18 +108,35 @@ namespace krabs {
...
+                        break;
+                    }
+                }
+                propertyLength = static_cast<ULONG>(((PBYTE)p) - propertyStart);
             }
             else if (propertyInfo.nonStructType.InType == TDH_INTYPE_ANSISTRING)
             {
-                // size of ansi string is length + 1 (null byte)
-                propertyLength = (ULONG)strlen((const char*)propertyStart) + 1;
+                const char* p = (const char*)propertyStart;
```
zacbrown: stylistically - I'd prefer to see us use `auto` here.

krabs/krabs/size_provider.hpp
```diff
@@ -105,18 +108,35 @@ namespace krabs {
...
+        // to the end of the record buffer.
+
         if (propertyInfo.nonStructType.OutType == TDH_OUTTYPE_STRING)
         {
             if (propertyInfo.nonStructType.InType == TDH_INTYPE_UNICODESTRING)
             {
-                // size of unicode string is length + 1 (null byte) * sizeof(wchar)
-                auto chars = (ULONG)wcslen((const wchar_t*)propertyStart) + 1;
-                propertyLength = chars * sizeof(wchar_t);
+                const wchar_t* p = (const wchar_t*)propertyStart;
```
zacbrown: Stylistically - prefer `auto` here.

krabs/krabs/size_provider.hpp
zacbrown: Prefer `auto` here

zacbrown: ditto line below as well

krabs/krabs/size_provider.hpp
zacbrown: ditto line below as well

krabs/krabs/parser.hpp
swannman: You're right, I missed the ampersand. Thanks!

swannman:
@aydany it looks like there are a few minor changes necessary before we can merge this in. Are you still willing to make those?

aydany:
I apologize for taking way too long to get back to this. I addressed the comment about style and added a unit test to cover the code changes. The test will fail when run against master, but passes here.

Please note that the test covers only the UNICODE string case. To cover the ANSI string case we need a valid ETW event that has an ANSI string as its last property. I happened to encounter the UNICODE case accidentally and I do not have an example that can be used for the ANSI case at this time.

Issue #42, opened 2018-04-06
Does not compile with /permissive-


jrave:
Latest versions of VS2017 have /permissive- as the default setting but unfortunately krabsetw does not compile with it enabled:

`1>c:\users\<foobar>\documents\visual studio 2017\projects\krabstesting\packages\krabsetw.1.0.11\lib\native\include\krabs\filtering\comparers.hpp(100): error C2760: syntax error: unexpected token '}', expected ';'`

`1>c:\users\<foobar>\documents\visual studio 2017\projects\krabstesting\packages\krabsetw.1.0.11\lib\native\include\krabs\filtering\comparers.hpp(101): note: see reference to class template instantiation 'krabs::predicates::comparers::iequal_to<T>' being compiled`

Used versions:
krabsetw: 1.0.11 nuget package
VS 2017 15.6.5

jrave:
Not entirely sure where you saw the large number of errors when switching to latest Windows 10 SDK. So had a quick look and made a PR at #43 . That one at least builds the examples and native tests cleanly.

PR #43, opened 2018-04-07
Fix errors when building with /permissive- (#42)

jrave:
This should address #42 

Tested that examples and tests build with:
* Current default of VS 2015 and SDK 8.1
* VS 2017 15.6.5, SDK 10.0.16299.0 and /permissive-

### DIFF
--- a/O365.Security.Native.ETW/O365.Security.Native.ETW.vcxproj
+++ b/O365.Security.Native.ETW/O365.Security.Native.ETW.vcxproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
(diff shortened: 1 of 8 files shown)


zacbrown:
Thanks for your help! Not sure why I got an error making these changes. Maybe just dirty build.

Could you add the `/permissive-` setting to the NativeExamples and krabstest projects? That'll just ensure the project continues to be compliant with `/permissive-` going forward.

zacbrown:
Hi,

Yes - let's go ahead and upgrade the project toolchain to VS2017 and Win10 SDK.

jrave:
Alright updated the projects and solution to VS 2017 and enabled `/permissive-` for the NativeExamples and  krabstests projects.

There is one caveat with this change:
Boost now throws the following messages: `Unknown compiler version - please run the configure tests and report the results`. Tests run fine so this is likely just an annoyance.

Tried updating to Boost 1.66.0 but that gave the following message along a similar line: `Info: Boost.Config is older than your compiler version - probably nothing bad will happen - but you may wish to look for an update Boost version.  Define BOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE to suppress this message.`. So it seemed not worth it to also drag a Boost update into this.

zacbrown:
Thanks!

I missed this the other day so I’ll be taking a look today.

PR #45, opened 2018-05-01
Add filtering by primitive types

mjeong92:
Add filtering by Boolean, Int16, Int32, and Int64.

### DIFF
--- a/O365.Security.Native.ETW/Filtering/Fluent.hpp
+++ b/O365.Security.Native.ETW/Filtering/Fluent.hpp
@@ -96,10 +96,23 @@ namespace O365 { namespace Security { namespace ETW {
 		/// </summary>
 		/// <param name="propertyName">the name of the property to match on</param>
 		/// <param name="value">the value of the property to match on</param>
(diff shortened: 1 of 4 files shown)


mjeong92:
The unit tests fail but I don't know why, when I test these filters on an active ETW trace the filtering works correctly, so I know it's not the issue with the Filters themselves, but something with the test classes.

zacbrown:
@mjeong92 - hmm, looks like it's failing to lookup the schema for that provider. You could try a different provider with int16/int32/int64 properties.

mjeong92:
So it has to be a real provider and I can't make up the fields to test them? Hmm.. I don't know where I can look up providers that might have those properties.

Why does the LogonType fail then? That one definitely exists on that provider and it is a Int32 property.

swannman:
@mjeong92 it looks like LogonType may be a `UINT32`. Could that be what is tripping us up here?

mjeong92:
I removed all the new filters excpet UInt32 because that's the only one I need and that's the only type I could immediately find for test cases.

PR #48, opened 2018-06-08
system_call_provider fix, and FileIo provider addition

bionicbee:
Hi,

It appears that the system_call_provider references the wrong GUID.  According to https://msdn.microsoft.com/en-us/library/windows/desktop/aa964786(v=vs.85).aspx it should be using the PerfInfo GUID and not the SystemTraceControl GUID.

I've also added providers for File IO.

### DIFF
--- a/krabs/krabs/kernel_providers.hpp
+++ b/krabs/krabs/kernel_providers.hpp
@@ -61,6 +61,22 @@ namespace krabs { namespace kernel {
         EVENT_TRACE_FLAG_DISK_IO_INIT,
         krabs::guids::disk_io);
 
+	/**
+	* <summary>A provider that enables file io events</summary>
+	*/
+	CREATE_CONVENIENCE_KERNEL_PROVIDER(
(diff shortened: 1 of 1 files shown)


krabs/krabs/kernel_providers.hpp
```diff
@@ -61,6 +61,22 @@ namespace krabs { namespace kernel {
...
+	/**
+	* <summary>A provider that enables file io events</summary>
+	*/
+	CREATE_CONVENIENCE_KERNEL_PROVIDER(
+		file_io_provider,
+		EVENT_TRACE_FLAG_FILE_IO,
+		krabs::guids::file_io);
+
+	/**
+	* <summary>A provider that enables file io events</summary>
```
zacbrown: This should probably say “provides file io init events”.

zacbrown: This should probably say “provides file io init events”.

PR #51, opened 2018-09-19
Address issues related to setting the C++ standard to C++17.

zacbrown:
Address issues related to setting the C++ standard to C++17. Includes an upgrade to boost 1.68 due to deprecation of previously used feature.


swannman:
Thanks Zac!

PR #54, opened 2018-09-27
Provide separate open() and process() session calls

aydany:
The current start() method for a trace session hides the EVENT_TRACE_LOGFILE that is modified by OpenTrace. This structure is occasionally needed so that the trace consumer can interpret the trace data correctly. In particular, it indicates if timestamps are in raw format or not.

Provided an alternative to start() that allows to open() a session and to process() the session in separate calls. Open returns EVENT_TRACE_LOGFILE and does everything start() does except for the call to ProcessTrace.

### DIFF
--- a/krabs/krabs/trace.hpp
+++ b/krabs/krabs/trace.hpp
@@ -135,6 +135,34 @@ namespace krabs {
         void stop();
 
         /**
+        * <summary>
+        * Opens a trace session.
+        * </summary>
+        * <example>
+        *    krabs::trace trace;
+        *    krabs::guid id(L"{A0C1853B-5C40-4B15-8766-3CF1C58F985A}");
+        *    provider<> powershell(id);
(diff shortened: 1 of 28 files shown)


krabs/krabs/trace.hpp
```diff
@@ -245,6 +273,22 @@ namespace krabs {
         manager.stop();
     }
 
+    template <typename T>
+    EVENT_TRACE_LOGFILE trace<T>::open()
+    {
+        eventsHandled_ = 0;
```
zacbrown: Is this in a scenario where you're doing a re-open? I'm unclear why only this variable needs to be reset?

aydany: I tried to provide functional equivalence to what start() does today. In essence, this PR provides open(); process(); as an alternative to start(). Since start() clears the eventsHandled_ counter, I did the same in open().

PR #68, opened 2019-02-13
Address code review feedback from @kallanreed

swannman:
@kallanreed this addresses most of your feedback from the previous PR

### DIFF
--- a/krabs/krabs/filtering/event_filter.hpp
+++ b/krabs/krabs/filtering/event_filter.hpp
@@ -100,7 +100,7 @@ namespace krabs {
     private:
         std::deque<provider_callback> callbacks_;
         filter_predicate predicate_{ nullptr };
-        std::vector<unsigned short> provider_filter_event_ids_{ 0 };
+        std::vector<unsigned short> provider_filter_event_ids_;
 
     private:
         template <typename T>
--- a/krabs/krabs/filtering/event_filter.hpp
+++ b/krabs/krabs/filtering/event_filter.hpp
@@ -6,7 +6,6 @@
 #include <evntcons.h>
 #include <functional>
 #include <deque>
-#include <map>
 
 #include "../compiler_check.hpp"
 
@@ -53,12 +52,22 @@ namespace krabs {
         /**
          * <summary>
          *   Constructs an event_filter that applies event id filtering by event_id
-         *	 which will be added to list of filtered event 
-         *   ids in ETW API. This way is more effective from performance point of view.
-         *	 Given optional predicate will be applied to ETW API filtered results
+         *   which will be added to list of filtered event ids in ETW API.
+         *   This way is more effective from performance point of view.
+         *   Given optional predicate will be applied to ETW API filtered results
          * </summary>
          */
         event_filter(unsigned short event_id, filter_predicate predicate=nullptr);
+
+        /**
+         * <summary>
+         *   Constructs an event_filter that applies event id filtering by event_id
+         *   which will be added to list of filtered event ids in ETW API.
+         *   This way is more effective from performance point of view.
+         *   Given optional predicate will be applied to ETW API filtered results
+         * </summary>
+         */
+        event_filter(std::vector<unsigned short> event_ids, filter_predicate predicate = nullptr);
 
         /**
          * <summary>
@@ -73,9 +82,9 @@ namespace krabs {
         template <typename U>
         void add_on_event_callback(const U &callback);
 
-        unsigned short provider_filter_event_id() const
+        std::vector<unsigned short> provider_filter_event_ids() const
         {
-            return provider_filter_event_id_;
+            return provider_filter_event_ids_;
         }
 
     private:
@@ -91,7 +100,7 @@ namespace krabs {
     private:
         std::deque<provider_callback> callbacks_;
         filter_predicate predicate_{ nullptr };
-        unsigned short provider_filter_event_id_{ 0 };
+        std::vector<unsigned short> provider_filter_event_ids_{ 0 };
 
     private:
         template <typename T>
@@ -107,8 +116,13 @@ namespace krabs {
     : predicate_(predicate)
     {}
 
-	inline event_filter::event_filter(unsigned short event_id, filter_predicate predicate/*=nullptr*/)
-    : provider_filter_event_id_(event_id),
+    inline event_filter::event_filter(std::vector<unsigned short> event_ids, filter_predicate predicate/*=nullptr*/)
+    : provider_filter_event_ids_(event_ids.begin(), event_ids.end()),
+      predicate_(predicate)
+    {}
+
+    inline event_filter::event_filter(unsigned short event_id, filter_predicate predicate/*=nullptr*/)
+    : provider_filter_event_ids_{ event_id },
       predicate_(predicate)
     {}
 
--- a/krabs/krabs/filtering/event_filter.hpp
+++ b/krabs/krabs/filtering/event_filter.hpp
@@ -82,7 +82,7 @@ namespace krabs {
         template <typename U>
         void add_on_event_callback(const U &callback);
 
-        std::vector<unsigned short> provider_filter_event_ids() const
+        const std::vector<unsigned short>& provider_filter_event_ids() const
         {
             return provider_filter_event_ids_;
         }
@@ -117,7 +117,7 @@ namespace krabs {
     {}
 
     inline event_filter::event_filter(std::vector<unsigned short> event_ids, filter_predicate predicate/*=nullptr*/)
-    : provider_filter_event_ids_(event_ids.begin(), event_ids.end()),
+    : provider_filter_event_ids_{ event_ids },
       predicate_(predicate)
     {}
 
--- a/krabs/krabs/ut.hpp
+++ b/krabs/krabs/ut.hpp
@@ -31,7 +31,7 @@ namespace krabs { namespace details {
 
         struct filter_settings{
             std::set<unsigned short> provider_filter_event_ids_;
-            filter_flags filter_flags_;
+            filter_flags filter_flags_{};
         };
 
         typedef std::map<krabs::guid, filter_settings> provider_filter_settings;
@@ -121,10 +121,6 @@ namespace krabs { namespace details {
         // for the same GUID are provided and request different provider flags.
         // TODO: Only forward the calls that are requested to each provider.
         for (auto &provider : trace.providers_) {
-            if (provider_flags.find(provider.get().guid_) != provider_flags.end()) {
-                provider_flags[provider.get().guid_].filter_flags_ = {};
-            }
-
             auto& settings = provider_flags[provider.get().guid_];
             settings.filter_flags_.level_       |= provider.get().level_;
             settings.filter_flags_.any_         |= provider.get().any_;
@@ -132,11 +128,9 @@ namespace krabs { namespace details {
             settings.filter_flags_.trace_flags_ |= provider.get().trace_flags_;
 
             for (const auto& filter : provider.get().filters_) {
-                if (filter.provider_filter_event_ids().size() > 0) {
-                    //native id existing, set native filters
-                    auto& provider_filter_event_ids = settings.provider_filter_event_ids_;
-                    provider_filter_event_ids.insert(filter.provider_filter_event_ids().begin(), filter.provider_filter_event_ids().end());
-                }
+                settings.provider_filter_event_ids_.insert(
+                    filter.provider_filter_event_ids().begin(),
+                    filter.provider_filter_event_ids().end());
             }
         }
 
--- a/krabs/krabs/ut.hpp
+++ b/krabs/krabs/ut.hpp
@@ -167,12 +167,8 @@ namespace krabs { namespace details {
                 filterEventIds->Count = static_cast<USHORT>(provider.second.provider_filter_event_ids_.size());
 
                 auto index = 0;
-                std::set<unsigned short>::iterator provider_filter_iterator;
-                for (provider_filter_iterator = provider.second.provider_filter_event_ids_.begin();
-                     provider_filter_iterator != provider.second.provider_filter_event_ids_.end();
-                     ++provider_filter_iterator)
-                {
-                    filterEventIds->Events[index] = *provider_filter_iterator;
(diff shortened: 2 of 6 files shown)


swannman:
cc @issaharnoam for review

krabs/krabs/ut.hpp
```diff
@@ -113,19 +122,19 @@ namespace krabs { namespace details {
...
             if (provider_flags.find(provider.get().guid_) != provider_flags.end()) {
-                provider_flags[provider.get().guid_].flags_tuple_ = std::make_tuple<UCHAR, ULONGLONG, ULONGLONG, UCHAR> (0, 0, 0, 0);
+                provider_flags[provider.get().guid_].filter_flags_ = {};
             }
 
-            std::get<0>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().level_;
-            std::get<1>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().any_;
-            std::get<2>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().all_;
-            std::get<3>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().trace_flags_;
+            provider_flags[provider.get().guid_].filter_flags_.level_ |= provider.get().level_;
```
swannman: @kallanreed you had a comment here - 

```Do the lookup once and store it in a reference variable. That way you don't need to do the lookup in the map multiple times. Also it would let you get rid of the for loop below (for event_id filtering). If you end up using the std::set instead of the vector, then you shouldn't need all that looping anyway (use std::copy with an inserter)```

...could you elaborate? I didn't quite understand the suggestion.

Thanks!

File: krabs/krabs/filtering/event_filter.hpp [hunk 2]
```diff
@@ -91,7 +100,7 @@ namespace krabs {
     private:
         std::deque<provider_callback> callbacks_;
         filter_predicate predicate_{ nullptr };
-        unsigned short provider_filter_event_id_{ 0 };
+        std::vector<unsigned short> provider_filter_event_ids_{ 0 };
```
swannman: remove `{0}` here

krabs/krabs/ut.hpp
```diff
@@ -152,14 +160,22 @@ namespace krabs { namespace details {
...
-                auto filterEventIds = reinterpret_cast<PEVENT_FILTER_EVENT_ID>(filterMemoryPtr.get());
+                auto filterEventIds = reinterpret_cast<PEVENT_FILTER_EVENT_ID>(&(filterMemoryPtr[0]));
                 filterEventIds->FilterIn = TRUE;
                 filterEventIds->Count = static_cast<USHORT>(provider.second.provider_filter_event_ids_.size());
-                for (int index = 0; index < filterEventIds->Count; ++index) {
-                    filterEventIds->Events[index] = provider.second.provider_filter_event_ids_[index];
+
+                auto index = 0;
+                std::set<unsigned short>::iterator provider_filter_iterator;
+                for (provider_filter_iterator = provider.second.provider_filter_event_ids_.begin();
```
zacbrown: An alternative way to do this would be to use the [range-based for-loop](https://en.cppreference.com/w/cpp/language/range-for) that is available to iterators:

```suggestion
                for (auto filter : provider.second.provider_filter_event_ids_) {
                     filterEventIds->Events[index] = filter;
                     index++;
                }
```

**Note: I didn't test that code, that's just a rough example.**

krabs/krabs/ut.hpp
kylereedmsft: Something like this
```
auto& flags = provider_flags[provider.get().guid_];
std::get<1>(flags.flags_tuple_) |= provider.get().any_;
...
```
Even better if you get rid of that `std::variant`

krabs/krabs/ut.hpp
```diff
@@ -113,19 +122,19 @@ namespace krabs { namespace details {
...
+            provider_flags[provider.get().guid_].filter_flags_.all_ |= provider.get().all_;
+            provider_flags[provider.get().guid_].filter_flags_.trace_flags_ |= provider.get().trace_flags_;
 
             for (const auto& filter : provider.get().filters_) {
-                if (filter.provider_filter_event_id() > 0) {
+                if (filter.provider_filter_event_ids().size() > 0) {
                     //native id existing, set native filters
                     auto& provider_filter_event_ids = provider_flags[provider.get().guid_].provider_filter_event_ids_;
-                    provider_filter_event_ids.push_back(filter.provider_filter_event_id()); 
+                    provider_filter_event_ids.insert(filter.provider_filter_event_ids().begin(), filter.provider_filter_event_ids().end());
```
swannman: when using provider filtering (`user_trace_003_no_predicates.cpp`) I fail the `std::vector` `_Verify_range` assertion `vector iterators in range are from different containers`.

I investigated but don't see why this is occurring. Am I missing something obvious? @zacbrown @kallanreed

krabs/krabs/filtering/event_filter.hpp
```diff
@@ -107,8 +116,13 @@ namespace krabs {
...
 
-	inline event_filter::event_filter(unsigned short event_id, filter_predicate predicate/*=nullptr*/)
-    : provider_filter_event_id_(event_id),
+    inline event_filter::event_filter(std::vector<unsigned short> event_ids, filter_predicate predicate/*=nullptr*/)
+    : provider_filter_event_ids_(event_ids.begin(), event_ids.end()),
+      predicate_(predicate)
+    {}
+
+    inline event_filter::event_filter(unsigned short event_id, filter_predicate predicate/*=nullptr*/)
+    : provider_filter_event_ids_{ event_id },
```
swannman: should this be `({ event_id })` instead of `{ event_id }`?

kylereedmsft: You forgot a possible `{{ event_id }}` (this language is a disaster).

```
template <int x>
struct Tag { };

class Foo
{
	std::vector<short> vs_;

public:								      // Foo f(5)
	Foo(short s, Tag<1>) : vs_(s) { }     // make a vector with 5 entries
	Foo(short s, Tag<2>) : vs_{ s } { }	  // make a vector with 1 entry, value 5
	Foo(short s, Tag<3>) : vs_({ s }) { } // make a vector with 1 entry, value 5
	Foo(short s, Tag<4>) : vs_{{ s }} { } // make a vector with 1 entry, value 5

	auto count() const { return vs_.size(); }
	auto first() const { return *vs_.begin(); }
};

int main() noexcept
{
	Foo f1(5, Tag<1>{});
	Foo f2(5, Tag<2>{});
	Foo f3(5, Tag<3>{});
	Foo f4(5, Tag<4>{});
	std::cout << f1.count() << ", " << f1.first() << std::endl;
	std::cout << f2.count() << ", " << f2.first() << std::endl;
	std::cout << f3.count() << ", " << f3.first() << std::endl;
	std::cout << f4.count() << ", " << f4.first() << std::endl;
}
```

kylereedmsft: Short answer, it's fine.

krabs/krabs/ut.hpp
```diff
@@ -113,19 +122,20 @@ namespace krabs { namespace details {
...
             if (provider_flags.find(provider.get().guid_) != provider_flags.end()) {
-                provider_flags[provider.get().guid_].flags_tuple_ = std::make_tuple<UCHAR, ULONGLONG, ULONGLONG, UCHAR> (0, 0, 0, 0);
+                provider_flags[provider.get().guid_].filter_flags_ = {};
             }
 
-            std::get<0>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().level_;
-            std::get<1>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().any_;
-            std::get<2>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().all_;
-            std::get<3>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().trace_flags_;
+            auto& flags = provider_flags[provider.get().guid_];
```
kylereedmsft: Maybe call this `settings` - I didn't see it in context and the type coming back is a `filter_settings`

krabs/krabs/filtering/event_filter.hpp
```diff
@@ -73,9 +82,9 @@ namespace krabs {
         template <typename U>
         void add_on_event_callback(const U &callback);
 
-        unsigned short provider_filter_event_id() const
+        std::vector<unsigned short> provider_filter_event_ids() const
```
kylereedmsft: Return a `const std::vector<unsigned short>&` or else you're making a copy of the vector.

krabs/krabs/ut.hpp
kylereedmsft: Le sigh (C++ is bad and it should feel bad), `provider_filter_event_ids()` is returning a copy so you're using iterators from two different instances (copies). I added a comment in the filter class.

krabs/krabs/filtering/event_filter.hpp
```diff
@@ -107,8 +116,13 @@ namespace krabs {
     : predicate_(predicate)
     {}
 
-	inline event_filter::event_filter(unsigned short event_id, filter_predicate predicate/*=nullptr*/)
-    : provider_filter_event_id_(event_id),
+    inline event_filter::event_filter(std::vector<unsigned short> event_ids, filter_predicate predicate/*=nullptr*/)
+    : provider_filter_event_ids_(event_ids.begin(), event_ids.end()),
```
kylereedmsft: In this case, prefer `provider_filter_event_ids_{ event_ids }` which will invoke the vector copy ctor.

krabs/krabs/ut.hpp
```diff
@@ -152,14 +160,22 @@ namespace krabs { namespace details {
 
                 //allocate + size of expected events in filter
                 DWORD size = FIELD_OFFSET(EVENT_FILTER_EVENT_ID, Events[provider.second.provider_filter_event_ids_.size()]);
-                filterMemoryPtr = std::make_unique<BYTE[]>(size);
+                std::vector<BYTE> filterMemoryPtr(size);
```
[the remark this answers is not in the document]
kylereedmsft: Also, call this like, `filterBuffer` or something. It's not a pointer anymore.

krabs/krabs/ut.hpp
```diff
@@ -113,19 +122,20 @@ namespace krabs { namespace details {
...
+            flags.filter_flags_.level_ |= provider.get().level_;
+            flags.filter_flags_.any_ |= provider.get().any_;
+            flags.filter_flags_.all_ |= provider.get().all_;
+            flags.filter_flags_.trace_flags_ |= provider.get().trace_flags_;
 
             for (const auto& filter : provider.get().filters_) {
-                if (filter.provider_filter_event_id() > 0) {
+                if (filter.provider_filter_event_ids().size() > 0) {
                     //native id existing, set native filters
                     auto& provider_filter_event_ids = provider_flags[provider.get().guid_].provider_filter_event_ids_;
```
kylereedmsft: `provider_flags[provider.get().guid_].provider_filter_event_ids_;` could be `flags.provider_filter_event_ids_`.

krabs/krabs/ut.hpp
```diff
@@ -137,11 +147,10 @@ namespace krabs { namespace details {
             parameters.SourceId = provider.first;
```
kylereedmsft: Very optional: you can use structured bindings to make all of the `provider.second` stuff below go away.
E.g.
```
	std::map<int, std::vector<int>> mapping;
	for (auto& [id, items] : mapping)
	{
		// id is the key
		// items is the value
	}
```
in this case `for (auto& [provider, settings] : provider_flags)` and then `provider.second` would be `settings` or whatever you name it.

krabs/krabs/ut.hpp
```diff
@@ -137,11 +147,10 @@ namespace krabs { namespace details {
...
             
             GUID guid = provider.first;
-            parameters.EnableProperty = std::get<3>(provider.second.flags_tuple_);
+            parameters.EnableProperty = provider.second.filter_flags_.trace_flags_;
             parameters.EnableFilterDesc = nullptr;
             parameters.FilterDescCount = 0;
             EVENT_FILTER_DESCRIPTOR filterDesc;
```
kylereedmsft: `EVENT_FILTER_DESCRIPTOR filterDesc{};` to agg-init to zero and then you don't need the `ZeroMemory` call below.

krabs/krabs/ut.hpp
```diff
@@ -113,19 +122,20 @@ namespace krabs { namespace details {
         // TODO: Only forward the calls that are requested to each provider.
         for (auto &provider : trace.providers_) {
             if (provider_flags.find(provider.get().guid_) != provider_flags.end()) {
```
kylereedmsft: Re-reading this, the `if` block here isn't needed at all. Just remove the whole thing.
To initialize the settings' `filter_flags_` member, add a agg-initializer to line 34 on the struct member `filter_flags filter_flags_{};` and you should be good.

krabs/krabs/ut.hpp
```diff
@@ -113,19 +122,20 @@ namespace krabs { namespace details {
...
-            std::get<3>(provider_flags[provider.get().guid_].flags_tuple_) |= provider.get().trace_flags_;
+            auto& settings = provider_flags[provider.get().guid_];
+            settings.filter_flags_.level_       |= provider.get().level_;
+            settings.filter_flags_.any_         |= provider.get().any_;
+            settings.filter_flags_.all_         |= provider.get().all_;
+            settings.filter_flags_.trace_flags_ |= provider.get().trace_flags_;
 
             for (const auto& filter : provider.get().filters_) {
-                if (filter.provider_filter_event_id() > 0) {
+                if (filter.provider_filter_event_ids().size() > 0) {
```
kylereedmsft: If you're willing to accept a tiny bit of overhead (it's only on start up anyway and it's really not much) you could get rid of this `if` block too and replace it with:
```
settings.provider_filter_event_ids.insert(
    filter.provider_filter_event_ids().begin(),
    filter.provider_filter_event_ids().end());
```

swannman:
Thanks @kylereedmsft, updated with your latest feedback.

@zacbrown @issaharnoam any feedback before I merge?

zacbrown:
Looks good to me @swannman .

PR #75, opened 2019-05-30
Added <vector> include to event_filter.hpp to make it compile on VS2019

bionicbeagle:
event_filter.hpp uses `std::vector` but does not contain `#include <vector>` 

This presumably worked previously because one of the existing includes transitively included vector but that is apparently no longer the case.


swannman:
@bionicbeagle thanks for submitting this!

PR #82, opened 2019-08-30
Catch and swallow destructor exceptions

swannman:
We have observed a crash in some environments with the following callstack:

callstack
We believe this is occurring during agent shutdown after another process in the environment has attempted to "clean up" the machine state by removing the Krabs user trace session. The `UserTrace` destructor calls `Stop()` which may throw if the trace was not found (see `error_check_common_conditions`). If an exception is thrown in the destructor, the process may exit via `std::terminate` ([ref](https://en.cppreference.com/w/cpp/error/terminate), [ref](https://akrzemi1.wordpress.com/2011/09/21/destructors-that-throw/)).

Our proposed fix is to `try/catch` around the call to `Stop()` in the destructor to prevent exceptions from leaking out of the destructor. We apply the same remedy to `UserTrace` and `KernelTrace` as the condition may occur in both destructors.

### DIFF
--- a/O365.Security.Native.ETW/UserTrace.hpp
+++ b/O365.Security.Native.ETW/UserTrace.hpp
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
 
     inline UserTrace::~UserTrace()
     {
-        if (disposed_) {
+        if (disposed_)
+        {
             return;
         }
 
-        Stop();
+        try
+        {
+            Stop();
+        }
+        catch (...)
+        {
+            // Stop may throw if the trace has been removed by another process.
+            // We catch and swallow the exception here to avoid an exception from leaking
+            // out of the destructor, which may cause a crash.
+        }
+
         disposed_ = true;
     }
 
--- a/O365.Security.Native.ETW/UserTrace.hpp
+++ b/O365.Security.Native.ETW/UserTrace.hpp
@@ -122,17 +122,14 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
 
     inline UserTrace::~UserTrace()
     {
-        if (disposed_)
-        {
+        if (disposed_) {
             return;
         }
 
-        try
-        {
+        try {
             Stop();
         }
-        catch (...)
-        {
+        catch (...) {
             // Stop may throw if the trace has been removed by another process.
             // We catch and swallow the exception here to avoid an exception from leaking
             // out of the destructor, which may cause a crash.
--- a/O365.Security.Native.ETW.Debug.nuspec
+++ b/O365.Security.Native.ETW.Debug.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Native.ETW.Debug</id>
-        <version>2.0.1</version>
+        <version>2.0.2</version>
         <title>Microsoft.O365.Security.Native.ETW Debug - managed wrappers for krabsetw</title>
         <authors>Microsoft</authors>
         <owners>Microsoft</owners>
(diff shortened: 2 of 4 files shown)


swannman:
cc @zacbrown and @kallanreed for feedback

O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
 
     inline UserTrace::~UserTrace()
     {
-        if (disposed_) {
+        if (disposed_)
```
zacbrown:
```suggestion
        if (disposed_) {
```

O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
+        {
```
zacbrown:
```suggestion
```

O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
             return;
         }
 
-        Stop();
+        try
```
zacbrown:
```suggestion
        try {
```

O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
+        {
```
zacbrown:
```suggestion
```

O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
+            Stop();
+        }
+        catch (...)
+        {
```
zacbrown:
```suggestion
```

O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
             return;
         }
 
-        Stop();
+        try
+        {
+            Stop();
+        }
+        catch (...)
```
zacbrown:
```suggestion
        catch (...) {
```

O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -122,11 +122,22 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
+        {
+            // Stop may throw if the trace has been removed by another process.
+            // We catch and swallow the exception here to avoid an exception from leaking
```
zacbrown: A blanket `catch (...)` makes me nervous. Given we know what the exception type could be that is problematic, it's probably better to catch those specific exceptions and let the remainder surface. Otherwise, you may be hiding/obscuring a more serious problem.

For example:

```
try {
    Stop();
}
catch (const krabs::no_trace_sessions_remaining&) {
    // comment about not throwing in this scenario
}
```

swannman:
Hitting pause on this -- we've made a change in a different area that appears to resolve this issue.

(Note that it's challenging to catch a specific exception in this change since the specific exception type/error code is not bubbled up to the caller.)

jdu2600:
I *think* that you just made a change that masked this behaviour -> but I suspect that the root cause is still there. Specifically unregister_trace() appears in two places - and the second would return 'not found' and cause the exception.  Currently there is no code path to the second call, but there is no code path to CloseTrace() either - a potential handle leak?.

trace_manager<T>::stop() calls unregister_trace(), setting trace_.sessionHandle_ to INVALID_PROCESSTRACE_HANDLE, and calls  stop_trace().
stop_trace() wants to call unregister_trace() and then CloseTrace(), but currently does nothing because of INVALID_PROCESSTRACE_HANDLE being set...

I stumbled over this while trying to add a feature.  PR with a fix on its way.

PR #90, opened 2020-01-01
  .stop() should always call ControlTrace(STOP) - even if .open()/.start() has not been called

jdu2600:
In C# I wanted to be able to call .stop() to clean up orphan traces - without having to start processing events.

Originally I planned to add C++/CLI support for .open() which should(?) have worked. But when I dug into it, it seemed neater to just always (blindly) call STOP and swallow any 'not found' status codes. Thoughts?

~~I then also noticed a superfluous call to unregister_trace() and that CloseTrace() was *never* being called.  (See comment on https://github.com/microsoft/krabsetw/pull/82)  So I've proposed a fix for this logic.~~

And finally, I propose renaming unregister_trace() ~~to~~ *and* stop_trace() ~~and vice versa~~.
unregister_trace was calling ControlTrace(STOP) - which is the converse operation to start_trace() *so rename to stop_trace()*
stop_trace was calling CloseTrace() on the session handle - which is the converse operation to ~~register_trace()~~ *open_trace(), so rename to close_trace()*

### DIFF
--- a/krabs/krabs/etw.hpp
+++ b/krabs/krabs/etw.hpp
@@ -267,18 +267,18 @@ namespace krabs { namespace details {
     template <typename T>
     EVENT_TRACE_PROPERTIES trace_manager<T>::query_trace()
     {
-		trace_info info = fill_trace_info();
-		ULONG status = ControlTrace(
-			NULL,
-			trace_.name_.c_str(),
-			&info.properties,
-			EVENT_TRACE_CONTROL_QUERY);
+        trace_info info = fill_trace_info();
+        ULONG status = ControlTrace(
+            NULL,
+            trace_.name_.c_str(),
+            &info.properties,
+            EVENT_TRACE_CONTROL_QUERY);
 
-		if (status != ERROR_WMI_INSTANCE_NOT_FOUND) {
-			error_check_common_conditions(status);
+        if (status != ERROR_WMI_INSTANCE_NOT_FOUND) {
+            error_check_common_conditions(status);
 
             return info.properties;
-		}
+        }
 
         return { };
     }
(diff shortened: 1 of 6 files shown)


krabs/krabs/etw.hpp
```diff
@@ -337,10 +335,9 @@ namespace krabs { namespace details {
     }
 
     template <typename T>
-    void trace_manager<T>::stop_trace()
+    void trace_manager<T>::unregister_trace()
```
swannman: looks like `trace_.registrationHandle_` is no longer being checked or assigned after this change. is that deliberate?

jdu2600: So... it wasn't at the time....  But it is now.

The registrationHandle is only used by enable_providers which is only ever called immediately after register_trace.  (So it could probably be refactored out entirely).

query_stats actually superfluously passes the registrationHandle to ControlTrace() but it also passes the trace name. And this functions ignores the handle in this case.

I'm going to update query_stats so that it also can be called before open/start. 
Currently I plan to swallow the 'trace not found' case - and return a 0-filled stats structure. 
You'll be able to check stats.BufferCount to determine if a trace is running.

This will also handle the previous edge case when krabs would throw an unknown exception if query_stats is called after ControlTrace(STOP)  e.g. logman stop MyTrace -ets

krabs/krabs/etw.hpp
```diff
@@ -252,37 +250,35 @@ namespace krabs { namespace details {
...
     {
-        if (trace_.registrationHandle_ != INVALID_PROCESSTRACE_HANDLE)
-        {
-            trace_info info = fill_trace_info();
-            ULONG status = ControlTrace(
-                trace_.registrationHandle_,
-                trace_.name_.c_str(),
-                &info.properties,
-                EVENT_TRACE_CONTROL_QUERY);
+		trace_info info = fill_trace_info();
```
swannman: nit - please convert tabs to spaces here and elsewhere

jdu2600: Done. And fixed my editor settings.

jdu2600:
Rebased.

PR #95, opened 2020-01-16
deprecate process()

jdu2600:
On second though, process() is a bit redundant and could be deprecated.

### DIFF
--- a/Microsoft.O365.Security.Native.ETW/KernelTrace.hpp
+++ b/Microsoft.O365.Security.Native.ETW/KernelTrace.hpp
@@ -100,28 +100,13 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         ///     // ...
         ///     trace.Open();
         ///     // ...
-        ///     trace.Process();
-        /// </example>
-        virtual void Open();
-
-        /// <summary>
-        /// Initiates the event processing loop for an open trace session.
-        /// </summary>
-        /// <example>
-        ///     var trace = new KernelTrace();
-        ///     // ...
-        ///     trace.Open();
-        ///     // ...
-        ///     trace.Process();
+        ///     trace.Start();
         /// </example>
         /// <remarks>
-        /// This function is a blocking call. Whichever thread calls Start() is effectively
-        /// donating itself to the ETW subsystem as the processing thread for events.
-        ///
-        /// A side effect of this is that it is expected that Stop() will be called on
-        /// a different thread.
+        /// This is an optional call before Start() if you need the trace
+        /// registered with the ETW subsystem before you start processing events.
         /// </remarks>
-        virtual void Process();
+        virtual void Open();
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


krabs/krabs/etw.hpp
```diff
@@ -158,7 +151,9 @@ namespace krabs { namespace details {
     template <typename T>
     void trace_manager<T>::start()
     {
-        (void)open();
+        if(trace_.sessionHandle_ == INVALID_PROCESSTRACE_HANDLE) {
```
swannman: nit: space between `if` and the opening paren

Microsoft.O365.Security.Native.ETW/UserTrace.hpp
```diff
@@ -108,28 +108,13 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
         /// <remarks>
-        /// This function is a blocking call. Whichever thread calls Start() is effectively
-        /// donating itself to the ETW subsystem as the processing thread for events.
-        ///
-        /// A side effect of this is that it is expected that Stop() will be called on
-        /// a different thread.
+        /// This is an optional call before Start() if you need the trace
+        /// registered with the ETW subsystem before you start processing events.
         /// </remarks>
-        virtual void Process();
```
swannman: since this is a public API, should we keep the method but mark it deprecated or obsolete instead of removing it?

Microsoft.O365.Security.Native.ETW/KernelTrace.hpp
```diff
@@ -100,28 +100,13 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
         /// <remarks>
-        /// This function is a blocking call. Whichever thread calls Start() is effectively
-        /// donating itself to the ETW subsystem as the processing thread for events.
-        ///
-        /// A side effect of this is that it is expected that Stop() will be called on
-        /// a different thread.
+        /// This is an optional call before Start() if you need the trace
+        /// registered with the ETW subsystem before you start processing events.
         /// </remarks>
-        virtual void Process();
```
swannman: since this is a public API, should we keep it but mark it deprecated or obsolete instead of removing it?

Microsoft.O365.Security.Native.ETW/UserTrace.hpp
jdu2600: It only became a public API 2 days ago.  :-)

jdu2600: Deprecating the C++/CLI versions when we only just added them felt a bit weird?
It's highly unlikely that anyone is using them yet.

swannman: You're totally right. I saw that a partner team was calling `process()` but they're using the C++ API directly, not the C++/CLI wrapper. My mistake!

Issue #61, opened 2018-11-11
Possible synchronization issue in schema_locator


wmatw:
Hello, in the file `schema_locator.hpp`  line 145 about the following function:
```
inline const PTRACE_EVENT_INFO schema_locator::get_event_schema(const EVENT_RECORD& record)
    {
        // check the cache
        auto key = schema_key(record);
        auto& buffer = cache_[key];

        // return if there's a cache hit
        if(buffer) return (PTRACE_EVENT_INFO)(&buffer[0]);

        auto temp = get_event_schema_from_tdh(record);

        // multiple threads may end up trying to save their
        // temp objects to the cache so we need to lock
        // so only the first entry should be cached.

        if(!buffer)
        {
            scope_lock l(sync_);
            if (!buffer) buffer.swap(temp);
        }

        return (PTRACE_EVENT_INFO)(&buffer[0]);
    }
```
The access to the unordered_map is not protected by synchronization mechanism at `auto& buffer = cache_[key];` an insertion can occur and that can result in a rehashing of the container, then all iterators will be invalidated.
If another thread is reading the container at the same exact moment, the behavior will be undefined. I think that using a "singleton double checked locking pattern" for insertion is not enough, a full locking is required to completely eliminate the threat...
Mathieu

zacbrown:
Hi Mathieu,
?
krabsetw runs entirely in a single thread. Do you have a repro that demonstrates this race condition? I’m not sure how it would be possible to encounter given that schema_locator is only expected to be used in the ETW callback which is single threaded per trace session.

Multiple trace sessions don’t share the schema locator and must build individual caches.

Thanks,

Zac

wmatw:
I think the code is not working as you expected:
See `schema_locator.hpp` line 137
```
private:
        static schema_locator singleton_;
```
if you are creating multiple sessions on different threads (classic when you are using both kernel and user providers by example - see example: kernel_and_user_trace_001.cpp ) every ETW callback called on a different thread will access **the same** singleton.
If we can remove the dangerous (and unnecessary, I think) reset function from this class, we can safely change the storage class of the cache to be ` static thread_local`, in that case, we can remove the ugly singleton, have a real lock free cache and BONUS the OS will deallocate the memory when the thread is gone (no need to have the reset function anyway).
If you support my idea I can make a PR ...
Mathieu

zacbrown:
Ah I see now. I had thought that cache was `thread_local`.

[...]

PR #96, opened 2020-01-17
[bugfix] thread-safe schema_locator (#61)

jdu2600:
I had a go at implementing a per-thread schema_locator using TlsAlloc().

thread_local doesn't seem to work for managed code (due to the differing definition of thread).


swannman:
Thanks for doing this! This looks reasonable to me -- @zacbrown @kylereedmsft any feedback?

krabs/krabs/schema_locator.hpp
```diff
@@ -124,18 +124,31 @@ namespace krabs {
...
-         * Get the schema_locator singleton instance.
+         * Get the schema_locator instance for this thread.
          * WARNING: if compiled into managed code, do NOT get a reference
          * to this in a static initializer or it will cause loader lock.
          */
-        static schema_locator& get_instance()
+        static schema_locator& get_thread_instance()
         {
-            return singleton_;
+            if (!index_)
```
kallanreed:
```suggestion
            if (index_ == 0)
```

krabs/krabs/schema_locator.hpp
```diff
@@ -124,18 +124,31 @@ namespace krabs {
...
+            {
+                scope_lock l(sync_);
+                if (!index_) index_ = TlsAlloc();
```
kallanreed:
```suggestion
                if (index_ == 0) index_ = TlsAlloc();
```

krabs/krabs.cpp
```diff
@@ -4,12 +4,13 @@
...
-// cache static variable. Due to issues with C++14 magic
+// cache static variables. Due to issues with C++14 magic
 // static and C++/CLI, we're unable to rely solely on
 // using magic static.
 
 // DO NOT ADD THINGS TO THIS FILE
 
 namespace krabs {
-    schema_locator schema_locator::singleton_;
+    DWORD schema_locator::index_;
```
kallanreed: I know it's actually zero-init, but I don't like seeing "uninitialized" variables getting read.
```suggestion
    DWORD schema_locator::index_ = 0;
```

krabs/krabs/schema_locator.hpp
```diff
@@ -124,18 +124,34 @@ namespace krabs {
...
+                if (index_ == 0) index_ = TlsAlloc();
+            }
+
+            if (index_ == TLS_OUT_OF_INDEXES)
+                throw std::runtime_error("TLS_OUT_OF_INDEXES");
+
+            auto tls_locator = static_cast<schema_locator*>(TlsGetValue(index_));
+            if (tls_locator == nullptr)
+            {
+                tls_locator = new schema_locator;
```
kallanreed: Sorry I messed this the first time, but this is leaked. The Tls block is freed for you, but this is heap allocated and never freed. You’ll need to figure out the callback that is called before the thread dies so you can call delete in the pointer stored here.

kallanreed: Maybe you can make and call a cleanup function in the per-thread that calls the function which pumps ETW. When the blocking ETW function returns, maybe you can call to reset or something then.

krabs/krabs.cpp
```diff
@@ -4,12 +4,13 @@
...
 // static and C++/CLI, we're unable to rely solely on
 // using magic static.
 
 // DO NOT ADD THINGS TO THIS FILE
 
 namespace krabs {
-    schema_locator schema_locator::singleton_;
-}
\ No newline at end of file
+    DWORD schema_locator::index_ = 0;
```
zacbrown: @kallanreed's C++ is most definitely better than mine these days, but this seems like a reasonable opportunity to use the `thread_local` storage specifier? I think it'd also allow you to get around having to explicitly use the TlsAlloc/TlsSetValue functions, giving the code a more natural feel.

I defer to @kallanreed in case there are gotchas or bears here. It just seemed like a natural place to use it.

More documentation: https://en.cppreference.com/w/cpp/language/storage_duration

kallanreed: Using the built-in thread local would be preferred so that you wouldn’t need to manage the locator cache memory at all - provided that works properly with managed c++. Off the top of my head, I don’t remember if this header is part of the “managed” part of the assembly in HIDS.

One thing I just remembered and you’ll definitely want to benchmark: this may add a pinvoke call from managed runtime to native for TlsGetData. We definitely need to be careful about that because the thunk can be expensive. IIRC that’s part of the reason we had added the cache in the first place. When krabs is compiled into managed C++ the call to tdh_get_schema was expensive because it was pinvoked.

Some additional docs here: https://docs.microsoft.com/en-us/cpp/parallel/thread-local-storage-tls?view=vs-2019

https://docs.microsoft.com/en-us/cpp/cpp/storage-classes-cpp?view=vs-2019#thread_local

krabs/krabs/schema_locator.hpp
jdu2600: I'll look into cleaning up the allocation when the blocking function returns. 
This seems more appropriate than a DLL_THREAD_DETACH approach - since we aren't allocating per-thread, just per tracing thread.

krabs/krabs.cpp
jdu2600: I'll have another look at thread_local - but my initial attempt bumped into
> error C2483: 'krabs::schema_locator::singleton_': object with destructor or non-aggregate type cannot be declared with thread storage duration in managed code

re: performance
We can probably avoid the pinvoke in the single-threaded single trace case if that helps?
The alternative approach would be to create a per-trace schema_locator and inject it appropriately - but the comments indicate that this was non-trivial to achieve.

krabs/krabs/schema_locator.hpp
jdu2600: Hmm. Actually this needs some more thought. schema_locator can be used outside of the processing thread - e.g. half of the tests. :-)

jdu2600: I don't think that DLL_THREAD_DETACH is an option - since the C++ implementation is headers only.

Unless... we #define different implementations for C++ and C++/CLI wrapper...
We could even use the preferred/performant thread_local approach for C++ then.

jdu2600:
I'm not entirely happy with https://github.com/microsoft/krabsetw/pull/96/commits/0552eab77f6bb16fbb7717852598a9137809f9a3 - since I can still leak schema_locators on non-processing threads. 

I'm not entirely happy with https://github.com/microsoft/krabsetw/pull/96/commits/48dc00aa1a1879822a1fe9898d62c8f7d5a4bd32 - since I can still cause an AV if I'm using a schema_locator on a non-processing thread.

But neither of those two scenarios seem likely??

kallanreed:
I think this needs a little more thinking around design. I really don't like the ref counting and all that. It seems like it would be better to register an application-wide "locator factory" that would encapsulate the differences in the scenarios. I need to look at the source a little more to see what the best way to manage the lifetime would be.

What is the problem we're actually trying to address here? Is it just the exception caused by the unsynchronized cache or is there a compelling reason to use the TLS cache? I am really concerned about the PInvoke cost to use the Windows APIs. They are really expensive. Also, because this runs on managed "threads" in some scenarios, TLS may not even be what we want because managed threads are not the same as real threads - so we have to be careful about mixing the two.

The safest thing to do may be to figure out what it would take to inject the cache instead.

kallanreed:
What if we just add shared locking around the cache? It should be lightweight in most cases, and only have a lock contention in the cases where the cache is being written to - which is mainly on start up. This would solve the the AV from unsynchronized access to the cache_ and should still avoid the managed/native thunking we are trying to avoid.

jdu2600:
I think that locking is the right approach.
And if perfomance does turn out to be an issue than we can look into injecting the cache.

Static TLS was meant to be a simple & performant solution. Dynamic TLS is turning out to be neither. :-/

I'll have a look tonight after work.

kallanreed:
I looked over the code again to refresh my memory. The entire point of this cache was to to prevent managed to native calls on the hot path. tdh already caches under the hood, so this cache isn't saving the lookup, it's saving the call to `TdhGetEventInformation`.

Because the goal is to avoid making native calls, any synchronization that ends up calling something from Windows (as most do) will end up adding back all the overhead we were trying to prevent.
Critical sections, mutexes (and other OS primatives), and interlocks are all out due to calls into Windows. std::mutex and std::atomic are out because they aren't supported in C++ CLI - so we're in a bad situation w.r.t. tools we have available to address this problem.

Ideally, the locator instance would be per-trace because there's only supposed to be one thread servicing a trace at a time. The tricky part is how to thread access to the trace through the rest of the API surface.

jdu2600:
I've uploaded a rough proof of concept for how the schema_locator injection could work. Feedback on the approach appreciated.  
And any C++ tips appreciated - since this project is my first experience with the language.
(I duplicated a fair bit of code - is there a better way?)

I couldn't see any alternatives to adding a new type of callback that also returned a schema_locator reference. So this will require clients to update to the new (threadsafe) APIs if they are doing multiple traces.

The (multi-threaded) NativeExamples and krabstests work with the new (threadsafe) APIs - but I haven't ported the changes to the CLR version yet. 

Currently I've kept the non-threadsafe API unchanged (other than removing the now unnecessary lock) - but we may wish to consider marking it deprecated.

kallanreed:
I like the approach, will take a closer look. At first glance, I think making a "threadsafe_provider" would help with the code churn - also, (@swannman) I think we should version the krabs package, bump the major version and make a breaking API change.

krabs/krabs/schema_locator.hpp
```diff
@@ -94,16 +95,11 @@ namespace krabs {
 
     /**
      * <summary>
-     * Fetches and caches schemas from TDH. Also has a singleton
-     * instance. It is preferable to inject the locator, but the
-     * current krabs implementation isn't built in a way that
-     * allows the schema_locator to be injected in all of the
-     * places where Schemas or Parsers are constructed.
+     * Fetches and caches schemas from TDH.
```
kallanreed:
```suggestion
     * Fetches and caches schemas from TDH.
     * NOTE: this cache also reduces the number of managed to native transitions
     * when krabs is compiled into a managed assembly.
```

krabs/krabs/schema_locator.hpp
```diff
@@ -148,19 +143,9 @@ namespace krabs {
...
-        if(!buffer)
-        {
-            scope_lock l(sync_);
-            if (!buffer) buffer.swap(temp);
+        if (!buffer) {
+            auto temp = get_event_schema_from_tdh(record);
+            buffer.swap(temp);
         }
 
         return (PTRACE_EVENT_INFO)(&buffer[0]);
```
kallanreed:
```suggestion
        return (PTRACE_EVENT_INFO)(buffer.get());
```

krabs/krabs/testing/event_filter_proxy.hpp
```diff
@@ -71,4 +71,27 @@ namespace krabs { namespace testing {
         event_filter_.on_event(record);
     }
 
+    class threadsafe_event_filter_proxy {
+    public:
+        threadsafe_event_filter_proxy(krabs::threadsafe_event_filter& filter);
+        void push_event(const synth_record& record);
```
kallanreed: [nit] The rest of the project has & and * on the identifier, not the type - please update to be consistent.

krabs/krabs/kt.hpp
```diff
@@ -119,11 +119,11 @@ namespace krabs { namespace details {
 
     inline void kt::forward_events(
         const EVENT_RECORD &record,
-        const krabs::trace<krabs::details::kt> &trace)
```
kallanreed: We want to keep this `const` everywhere - making the locator const will fix that.

krabs/krabs/trace.hpp
```diff
@@ -233,6 +234,8 @@ namespace krabs {
 
         EVENT_TRACE_PROPERTIES properties_;
 
+        schema_locator schema_locator_;
```
kallanreed: Keep a trace_context instead, Make it mutable so we can keep everything const on the API.
```suggestion
        mutable trace_context context_;
```

krabs/krabs/trace.hpp
```diff
@@ -212,6 +212,7 @@ namespace krabs {
          */
         size_t buffers_processed() const;
 
+
```
kallanreed:
```suggestion
```

krabs/krabs/ut.hpp
```diff
@@ -74,7 +74,7 @@ namespace krabs { namespace details {
          */
         static void forward_events(
             const EVENT_RECORD &record,
-            const krabs::trace<krabs::details::ut> &trace);
+            krabs::trace<krabs::details::ut> &trace);
```
kallanreed:
```suggestion
            const krabs::trace<krabs::details::ut> &trace);
```

krabs/krabs/ut.hpp
```diff
@@ -192,11 +192,11 @@ namespace krabs { namespace details {
 
     inline void ut::forward_events(
         const EVENT_RECORD &record,
-        const krabs::trace<krabs::details::ut> &trace)
+        krabs::trace<krabs::details::ut> &trace)
```
kallanreed:
```suggestion
        const krabs::trace<krabs::details::ut> &trace)
```

krabs/krabs/kt.hpp
```diff
@@ -69,7 +69,7 @@ namespace krabs { namespace details {
          */
         static void forward_events(
             const EVENT_RECORD &record,
-            const krabs::trace<krabs::details::kt> &trace);
+            krabs::trace<krabs::details::kt> &trace);
```
kallanreed:
```suggestion
            const krabs::trace<krabs::details::kt> &trace);
```

krabs/krabs/kt.hpp
```diff
@@ -119,11 +119,11 @@ namespace krabs { namespace details {
 
     inline void kt::forward_events(
         const EVENT_RECORD &record,
-        const krabs::trace<krabs::details::kt> &trace)
+        krabs::trace<krabs::details::kt> &trace)
```
kallanreed:
```suggestion
        const krabs::trace<krabs::details::kt> &trace)
```

krabs/krabs/provider.hpp
```diff
@@ -38,6 +39,9 @@ namespace krabs {
     typedef void(*c_provider_callback)(const EVENT_RECORD &);
     typedef std::function<void(const EVENT_RECORD &)> provider_callback;
 
+    typedef void(*threadsafe_c_provider_callback)(const EVENT_RECORD&, schema_locator&);
```
kallanreed:
```suggestion
    typedef void(*threadsafe_c_provider_callback)(const EVENT_RECORD &, const trace_context &);
```

krabs/krabs/provider.hpp
```diff
@@ -38,6 +39,9 @@ namespace krabs {
...
+    typedef std::function<void(const EVENT_RECORD&, schema_locator&)> threadsafe_provider_callback;
```
kallanreed:
```suggestion
    typedef std::function<void(const EVENT_RECORD &, const trace_context  &)> threadsafe_provider_callback;
```

krabs/krabs/provider.hpp
```diff
@@ -10,6 +10,7 @@
```
kallanreed: This definitely feels like we should either split this into another class `threadsafe_provider` or just make this one thread-safe and force the new API in a major version update.
At the very least, a bunch of the code dupe could be cleaned up with a CRTP pattern.
If you want an example of how the CRTP code would look, let me know.

krabs/krabs/provider.hpp
```diff
@@ -353,6 +396,15 @@ namespace krabs {
             for (auto &filter : filters_) {
                 filter.on_event(record);
             }
+
+            for (auto& callback : threadsafe_callbacks_) {
```
kallanreed: Here and everywhere: * and & on the identifier.
```suggestion
            for (auto &callback : threadsafe_callbacks_) {
```

krabs/krabs/trace.hpp
kallanreed: Hmm, let's pass a `const trace_context &` instead of the schema_locator.
That way, if we need to thread other things through, we'll just add them to the context.
```suggestion
struct trace_context
{
    const schema_locator schema_locator_;
    /* Add additional trace context here. */
};
```

This will allow per-trace objects to be passed along on the callback which may be useful in the future if you have a single callback registered with multiple traces.

krabs/krabs/trace.hpp
```diff
@@ -252,6 +255,7 @@ namespace krabs {
     , sessionHandle_(INVALID_PROCESSTRACE_HANDLE)
     , eventsHandled_(0)
     , buffersRead_(0)
+    , schema_locator_()
```
kallanreed:
```suggestion
    , context_()
```

krabs/krabs/trace.hpp
```diff
@@ -263,6 +267,7 @@ namespace krabs {
...
```
kallanreed:
```suggestion
    , context_()
```

krabs/krabs/ut.hpp
```diff
@@ -192,11 +192,11 @@ namespace krabs { namespace details {
...
     {
         for (auto &provider : trace.providers_) {
             if (record.EventHeader.ProviderId == provider.get().guid_) {
-                provider.get().on_event(record);
+                provider.get().on_event(record, trace.schema_locator_);
```
kallanreed:
```suggestion
                provider.get().on_event(record, trace.context_);
```

krabs/krabs/kt.hpp
```diff
@@ -119,11 +119,11 @@ namespace krabs { namespace details {
...
     {
         for (auto &provider : trace.providers_) {
             if (provider.get().id() == record.EventHeader.ProviderId) {
-                provider.get().on_event(record);
+                provider.get().on_event(record, trace.schema_locator_);
```
kallanreed:
```suggestion
                provider.get().on_event(record, trace.context_);
```

krabs/krabs/schema_locator.hpp
```diff
@@ -135,7 +131,6 @@ namespace krabs {
 
     private:
         static schema_locator singleton_;
-        critical_section sync_;
         std::unordered_map<schema_key, std::unique_ptr<char[]>> cache_;
```
kallanreed: Make this mutable so the whole class can be used in const context.
Make the schema_locator::get_event_schema() const too.

```suggestion
        mutable std::unordered_map<schema_key, std::unique_ptr<char[]>> cache_;
```

krabs/krabs/schema.hpp
```diff
@@ -54,6 +54,8 @@ namespace krabs {
          */
         schema(const EVENT_RECORD &);
 
+        schema(const EVENT_RECORD &, krabs::schema_locator &);
```
kallanreed: You'll need to make all references to the locator `const`

krabs/krabs/provider.hpp
```diff
@@ -101,12 +115,15 @@ namespace krabs {
              *   Called when an event occurs, forwards to callbacks and filters.
              * </summary>
              */
-            void on_event(const EVENT_RECORD &record) const;
+            void on_event(const EVENT_RECORD &record, krabs::schema_locator &schema_locator) const;
```
kallanreed:
```suggestion
            void on_event(const EVENT_RECORD &record, krabs::trace_context  &context) const;
```

krabs/krabs/provider.hpp
```diff
@@ -344,7 +361,33 @@ namespace krabs {
...
+        }
+
+        template <typename T>
+        void base_provider<T>::add_threadsafe_filter(const threadsafe_event_filter& f)
+        {
+            threadsafe_filters_.push_back(f);
+        }
+
+        template <typename T>
+        void base_provider<T>::on_event(const EVENT_RECORD &record, krabs::schema_locator &schema_locator) const
```
kallanreed:
```suggestion
        void base_provider<T>::on_event(const EVENT_RECORD &record, krabs::treace_context &context) const
```

krabs/krabs/provider.hpp
```diff
@@ -353,6 +396,15 @@ namespace krabs {
             for (auto &filter : filters_) {
                 filter.on_event(record);
             }
+
+            for (auto& callback : threadsafe_callbacks_) {
+                callback(record, schema_locator);
```
kallanreed:
```suggestion
                callback(record, context);
```

krabs/krabs/provider.hpp
```diff
@@ -353,6 +396,15 @@ namespace krabs {
...
+            }
+
+            for (auto& filter : threadsafe_filters_) {
+                filter.on_event(record, schema_locator);
```
kallanreed:
```suggestion
                filter.on_event(record, context);
```

krabs/krabs/schema_locator.hpp
kallanreed: This important comment was lost in as part of the review:
 https://github.com/microsoft/krabsetw/pull/96/files/6ae1e9234571f370df03b8d8dbffa62419a21cc0#r371471476

krabs/krabs/schema_locator.hpp
jdu2600: TIL: C++ has a mutable keyword.  :-)

jdu2600:
Pushed c01a9cc. Feedback appreciated.
Interestingly this approach breaks the C++ API - but not the C++/CLI API.

So we could actually use different implementations for C++ and C++/CLI so that we don't break either API... 
i.e. a thread_local schema_locator for C++, and an injected schema_locator for C++/CLI.
Let me know if you'd like to see what this would look like. 

I'm happy either way.

krabs/krabs/filtering/predicates.hpp
```diff
@@ -121,7 +121,8 @@ namespace krabs { namespace predicates {
 
             bool operator()(const EVENT_RECORD &record) const
             {
-                krabs::schema schema(record);
+                krabs::schema_locator schema_locator;
```
jdu2600: :TODO: inject schema_locator into predicates also

krabs/krabs/provider.hpp
jdu2600: Went down the major version update / API breaking change path. 
Waiting to hear from @swannman if this is his preference.

swannman: Yes, I am okay with a breaking change + major version update.

krabs/krabs/schema.hpp
```diff
@@ -42,17 +42,18 @@ namespace krabs {
...
         /**
          * <summary>
-         * Constructs a schema from an event record instance.
+         * Constructs a schema from an event record instance
+         * using the provided schema_locator.
          * </summary>
          *
          * <example>
-         *   void event(const EVENT_RECORD &record)
+         *   void event(const EVENT_RECORD &record, const krabs::trace_context &trace_context)
```
kallanreed:
```suggestion
         *   void on_event(const EVENT_RECORD &record, const krabs::trace_context &trace_context)
```

kallanreed:
> Interestingly the injected schema_locator approach breaks the C++ API - but not the C++/CLI API.
@jdu2600 What do you mean by this? There shouldn't be anything breaking in the injected approach.

krabs/krabs/schema.hpp
swannman: I'm okay with checking this in to `master`. If we're making an API-breaking change then let's bump the major version, otherwise this can be a dot release.

jdu2600:
In short -
All of the C++ examples had to be updated.
All of the C# examples worked as is.

@kallanreed - I've modified the existing provider_callback type to also take the trace_context as a second parameter which breaks all existing C++ code that uses provider::add_on_event_callback(). 
 
**This is different to the initial poc version you looked at where I left this intact and instead added an additional threadsafe_provider_callback type and an add_on_event_threadsafe_callback() method.**

The C++/CLI user API doesn't change - since the callback there receives an EventRecord which already (internally) had a reference to a schema_locator. (I just changed this to the per-trace schema_locator instead of the singleton).

swannman:
@jdu2600 do you want to bump the major version on the .nuspec files?

jdu2600:
@swannman bumped the versions as best I could - I've never worked with .nuspec files before.

swannman:
Looks good! I will upload updated packages to nuget.org in a couple of weeks when I am back from travel.

Issue #101, opened 2020-02-22
Failure to Build Microsoft.O365.Security.Native.ETW from external Solution


pathtofile:
Some people like myself want to build Native KrabsETW themselves as part of their own buildchain, instead of relying on the prebuilt NuGet package. Some of the reasons could be:
* Use dev/experimental branches of KrabsETW
* Use a more up-to-date version of KrabsETW that is in NuGet (today the NuGet package is a few weeks older than `master`)

I attempted to Create my own VS Solution, add KrabsETW as a git submodule, and add the `Microsoft.O365.Security.Native.ETW` project to the solution.

 The build failed with two things:
* A missing explicit `AdditionalIneludeDirectories` for the `krabs` folder.
  * Fixed in PR here: https://github.com/microsoft/krabsetw/pull/100
* Failure to find Boost Nuget package
  * This is due to how NuGet and Git Submodules (fail to) work together. I don't consider this KrabsETW's problem to solve, so I made a workaround in my own solution)

For reference, project is here: https://github.com/pathtofile/SimpleETW
If https://github.com/microsoft/krabsetw/pull/100 was merged, then this use-case becomes viable

pathtofile:
Thanks for Merging https://github.com/microsoft/krabsetw/pull/100
Found one more this, Project is looking for a file in the wrong spot in a PreBuildEvent, I think: https://github.com/microsoft/krabsetw/pull/102

pathtofile:
Both got merged, so I consider this issue sloved, thanks @swannman and team!

PR #102, opened 2020-02-25
PreBuildEvent that was looking for 35MSSharedLib1024.snk in wrong folder

pathtofile:
Linked to my Issue related to building KrabsETW Native as part of my own external project: https://github.com/microsoft/krabsetw/issues/101

`$(SolutionDir)...` puts the file outside the codebase, so I'm assuming this was a mistake? `$(ProjectDir)...` seems to make more sence, as that puts it at the project root, where there is the file already existing in the codebase: https://github.com/microsoft/krabsetw/blob/master/35MSSharedLib1024.snk


swannman:
@pathtofile does the VS solution at `/krabs/krabs.sln` still build properly with this change?

pathtofile:
@swannman Yep, doesn't prevent `krabs/krabs.sln` for building as-per usual.

Issue #87, opened 2019-11-18
Parsing erros for fields at the end of the record


fs-granma:
Hi,

I've ran into some problems while consuming events from the Microsoft-Windows-WinINet provider ({43d1a55c-76d6-4f7e-995c-64c711e5cafe}) on Windows 10.

Sometimes the last properties in the record cannot be parsed.

Sometimes the problematic properties (e.g. Event id: 1008 Prop: StatusLine, Event id: 506 Prop: URL) seem to be zero-length, non-null-terminated strings at the end of the record (i.e. there is a record with zero bytes of data). These hit the assertion 'invariant: if we've exhausted our buffer, then we must've exhausted the properties as well'. It seems that the end of the buffer might not strictly mean the end of properties.

There are also integer types that either throw an exception ('Property length past end of property buffer') or hit the assertion.

I have not observed similar issues with other providers so far.

Please find attached a list of event ids and property names from the Microsoft-Windows-WinINet provider that I have observed causing problems. Note that these don't always cause problems, but some properties (StatusLine, URL) seem to have a high probability to hit the assertion.

--Markus
[Microsoft-Windows-WinINet_parsing_issues.txt](https://github.com/microsoft/krabsetw/files/3859294/Microsoft-Windows-WinINet_parsing_issues.txt)

fs-granma:
Hello,

Any ideas how to proceed with this?

I'm thinking that if it's the provider (Microsoft-Windows-WinINet in this case) that produces invalid data then the library should probably handle it somehow gracefully rather than asserting.

It also seems that the assertion 'invariant: if we've exhausted our buffer, then we must've exhausted the properties as well' should probably be somehow relaxed since it seems like a legitimate case of having a zero-length string at the end of the buffer.

Having integer types past the end of the buffer seems a bit suspicious though.

Thanks,
Markus

jrave:
Hello,

we are still looking for a solution to this and hope that you can provide some guidance for this. At the end there is a small sample program that reproduces the behavior. The easiest for us has been to use Edge on Win 10 to trigger the issue.

What it seems like to us that some properties come in pairs <Name> and <Name>Length. If <Name>Length == 0 then <Name> should not even be attempted to parse. Question is now whether there can be done something in krabsetw to help with this or whether this should be purely handled on the application side.

Best Regards,
Johannes

```cpp
#include <iostream>
#include "..\..\krabs\krabs.hpp"
#include "examples.h"

void user_trace_wininet::start()
{
[...]

PR #105, opened 2020-03-08
Wininet parsing

jrave:
Fixed some small issues that prevent successful parsing of events from the WinINet provider. This should address the issues in #87


krabs/krabs/parser.hpp
```diff
@@ -142,7 +142,9 @@ namespace krabs {
         assert((pBufferIndex_ <= pEndBuffer_ && pBufferIndex_ >= schema_.record_.UserData) &&
                "invariant: we should've already thrown for falling off the edge");
 
-        assert((pBufferIndex_ == pEndBuffer_ ? lastPropertyIndex_ == totalPropCount
+        // accept that last property can be omitted from buffer. this happens if last property
+        // is string but empty and the provider strips the null terminator
+        assert((pBufferIndex_ == pEndBuffer_ ? ((totalPropCount - lastPropertyIndex_) <= 1)
```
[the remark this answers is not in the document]
jrave: Interesting point. I admittedly have not yet come across events with two zero length strings yet but I think theoretically it might be possible. Would be totally okay with removing the assert completely. Should I do that then?

krabs/krabs/size_provider.hpp
```diff
@@ -70,6 +70,15 @@ namespace krabs {
             // length is a union that may refer to another field for a length
             // value. In that case, defer to TDH for the value otherwise
             // use the length value directly.
+
+            // Some providers set length == 8 for 32bit pointer values. But
+            // only put 4 bytes into the buffer. Force correct size for pointers.
+            if (propertyInfo.nonStructType.InType == TDH_INTYPE_POINTER &&
```
kallanreed: I want to understand this a little more. It seems like this could be very error prone depending on whether the logging process in 32 or 64. Did you find documentation about this? If not, we should look into this harder.

krabs/krabs/parser.hpp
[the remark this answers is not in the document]
kallanreed: Is there any indication in the schemas that certain fields may not appear in the record? That seems like a critical flaw if not. I would like to see if this behavior is documented somewhere.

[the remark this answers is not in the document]
jrave: This one is based on the comments in tdh.h: https://github.com/thnetii/windows-sdk-headers/blob/master/um/tdh.h#L124 For this there was also already a test in krabs (parse_unicode_string_should_work_when_unicode_string_property_is_last_and_not_null_terminated) to cover this. But it did not fully encapsulate the issue as it only tries to parse the last value. And then the assert does not trigger as nothing from the message buffer has been consumed. The new test shows that behavior.

krabs/krabs/size_provider.hpp
jrave: https://github.com/thnetii/windows-sdk-headers/blob/master/um/tdh.h#L164 seems to give a hint on this. E.g. it suggests that the header should be used to determine size. I can try to dig out the exact events I was observing this with if you want to investigate further. Just let me know.

jrave: krabs_case

So example in the screenshot. UserData (0x00000125a7e601c8) is shown in the memory view. And in the memory one can see that the pointer is 4 bytes. At byte 5 there is a string starting. But propertyInfo.length is 8. So this would cause krabs to give out corrupted data.

This is from provider {43D1A55C-76D6-4F7E-995C-64C711E5CAFE} and event id is 1179 . This is produced by using Edge on Win 10 1903.

swannman: This also looks related: https://docs.microsoft.com/en-us/windows/win32/api/tdh/nf-tdh-tdhformatproperty

`PointerSize`: The size of a pointer value in the event data. To get the size, access the `EVENT_RECORD.EventHeader.Flags` member. The pointer size is 4 bytes if the `EVENT_HEADER_FLAG_32_BIT_HEADER` flag is set; otherwise, it is 8 bytes if the `EVENT_HEADER_FLAG_64_BIT_HEADER` flag is set.

krabs/krabs/parser.hpp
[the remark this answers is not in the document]
swannman: There is an interesting code comment here: https://docs.microsoft.com/en-us/windows/win32/etw/using-tdhformatproperty-to-consume-event-data

`The length will be zero for variable length, null-terminated strings and structures.`

krabs/krabs/size_provider.hpp
kallanreed: If that's the case then instead of this condition, maybe we should have something along these lines:
```
if (is_pointer) { return record.EventHeader.Flags & EVENT_HEADER_FLAG_32_BIT_HEADER ? 4 : 8; }
```

swannman: Agreed, that is better for readability.

jrave: Updated the PR accordingly

krabs/krabs/parser.hpp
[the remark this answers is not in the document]
swannman: @jrave thanks for pointing out the comment from `tdh.h`:

> Note that some event providers do not correctly nul-terminate the last string field in the event. While this is technically invalid, event decoders may silently tolerate such behavior instead of rejecting the event as invalid.

Given that the Microsoft comment specifically calls out the "last string field in the event" I am okay with your proposed change.

Issue #109, opened 2020-03-12
Latest changes break compatibility with Windows 7


jrave:
Changes in #107 cause incompatibility of krabs on Windows 7. Specifically it is related to the usage of TraceQueryInformation API which is only supported on Windows 8 and newer: https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-tracequeryinformation

@jdu2600 : Do you think it would be possible to rework that change to not rely on this API?

jdu2600:
TraceSetInformation is a light wrapper around NtSetSystemInformation - which provides an alternate approach to supporting Windows 7.  

See https://geoffchappell.com/studies/windows/km/ntoskrnl/api/etw/tracesup/groupmask.htm

swannman:
Fixed by #113

PR #110, opened 2020-03-12
Fix for Windows 7 compatibility of group mask feature

jrave:
This should address #109

### DIFF
--- a/examples/NativeExamples/NativeExamples.vcxproj
+++ b/examples/NativeExamples/NativeExamples.vcxproj
@@ -87,7 +87,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>TYPEASSERT;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>WINVER=0x0601;_WIN32_WINNT=0x601;TYPEASSERT;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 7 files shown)


swannman:
@jrave could we try the `NtSetSystemInformation` approach that @jdu2600 recommended on #109?

jrave:
> Note that although EventTraceGroupMaskInformation is valid for querying in version 6.0 and higher, versions before 6.2 reject it for setting. The returned error code is ERROR_NOT_IMPLEMENTED.

That is from https://geoffchappell.com/studies/windows/km/ntoskrnl/api/etw/tracesup/groupmask.htm and means that we could use NtQuerySystemInformation to extract the group mask. However, then we would not be able to update it. So I am not sure that feature can be made to work on Windows 7 at all.

Therefore the graceful fallback here seems like a sane choice to me.

Proposal could be to add an exception to the `kernel_provider(const GUID& id, PERFINFO_MASK group_mask)` constructor that would throw on Win 7 to make it easier to catch issues due to this.

swannman:
Thanks @jrave. I like your suggestion - I would prefer a change that throws an exception when an API is called on an unsupported operating system (like Win 7) rather than introducing a separate codepath that can only be tested on a certain downlevel OS. We will never be confident that future changes in this area work on Win 7 without a manual test pass, which I'd like to avoid.

jdu2600:
Sorry for the earlier terse comment - I've been out of the 'office'.

Looks like non-trivial investigation would be required to determine if PERFINFO_MASK support is possible on Windows 7 - which isn't likely worth the effort. 

I've thrown up the alternate approach for the fix (https://github.com/microsoft/krabsetw/pull/113).
The code is simpler - but it relies on the native API...

swannman:
Fixed by #113

PR #113, opened 2020-03-17
Fixed Windows 7 support (#110)

jdu2600:
This is an alternate fix using the native API.

### DIFF
--- a/Microsoft.O365.Security.Native.ETW/KernelProvider.hpp
+++ b/Microsoft.O365.Security.Native.ETW/KernelProvider.hpp
@@ -40,6 +40,7 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         /// <param name="id">the guid of the kernel trace</param>
         /// <param name="mask">the group mask to set</param>
         /// <remarks>
+        /// Only supported on Windows 8 and newer.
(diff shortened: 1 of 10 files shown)


jrave:
Looking okay to me. Just thinking that it would be great to also add the defines WINVER=0x0601;_WIN32_WINNT=0x601; to make it a bit harder to use unavailable APIs going forward.

jrave:
On second thought I am wondering on whether this would actually work on Windows 7. According to https://geoffchappell.com/studies/windows/km/ntoskrnl/api/etw/tracesup/groupmask.htm : 

> Note that although EventTraceGroupMaskInformation is valid for querying in version 6.0 and higher, versions before 6.2 reject it for setting. The returned error code is ERROR_NOT_IMPLEMENTED.

Would that not mean that `kt::enable_providers` would always throw an exception on Windows 7, meaning you could not create any kernel trace providers on Windows 7. Or am I misunderstanding something here?

jdu2600:
No. I missed something. The call is now only made if a group_mask has been supplied.

jrave:
Thanks. I think the updated code has a chance of working.

jdu2600:
@jrave - let me know if anything isn't behaving as expected (on Windows 7).

PR #126, opened 2020-04-17
Remove Boost Dependency

turingcompl33t:
Re-implement four functors in `comparers.hpp` with standard library algorithms in place of Boost algorithms.

### DIFF
--- a/krabs/krabs/filtering/comparers.hpp
+++ b/krabs/krabs/filtering/comparers.hpp
@@ -8,9 +8,9 @@
 #pragma warning(disable: 4800) // bool warning in boost when _MANAGED flag is set
 #endif
 
+#include <algorithm>
+
 #include "../compiler_check.hpp"
-#include <boost/algorithm/string.hpp>
-#include <boost/range/iterator_range_core.hpp>
 
 #if(_MANAGED)
 #pragma warning(pop)
@@ -32,10 +32,7 @@ namespace krabs { namespace predicates {
             template <typename Iter1, typename Iter2>
             bool operator()(Iter1 begin1, Iter1 end1, Iter2 begin2, Iter2 end2) const
             {
-                auto r1 = boost::make_iterator_range(begin1, end1);
-                auto r2 = boost::make_iterator_range(begin2, end2);
-
-                return boost::equals(r1, r2, Comparer());
+                return std::equal(begin1, end1, begin2, end2, Comparer());
             }
         };
 
@@ -48,10 +45,7 @@ namespace krabs { namespace predicates {
             template <typename Iter1, typename Iter2>
             bool operator()(Iter1 begin1, Iter1 end1, Iter2 begin2, Iter2 end2) const
             {
-                auto r1 = boost::make_iterator_range(begin1, end1);
-                auto r2 = boost::make_iterator_range(begin2, end2);
-
-                return boost::contains(r1, r2, Comparer());
+                return std::search(begin1, end1, begin2, end2, Comparer()) != end1;
             }
         };
 
@@ -64,10 +58,20 @@ namespace krabs { namespace predicates {
             template <typename Iter1, typename Iter2>
             bool operator()(Iter1 begin1, Iter1 end1, Iter2 begin2, Iter2 end2) const
             {
-                auto r1 = boost::make_iterator_range(begin1, end1);
-                auto r2 = boost::make_iterator_range(begin2, end2);
+                const auto dist1 = std::distance(begin1, end1);
+                const auto dist2 = std::distance(begin2, end2);
 
-                return boost::starts_with(r1, r2, Comparer());
+                // always starts with empty range
+                if (dist2 == 0)
+                    return true;
+
+                if (dist2 > dist1)
+                    return false;
+
+                auto prefix_end = begin1;
+                std::advance(prefix_end, dist2);
+
+                return std::equal(begin1, prefix_end, begin2, end2, Comparer());
             }
         };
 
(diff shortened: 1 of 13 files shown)


turingcompl33t:
I wrote a quick test suite to ensure that the semantics of the algorithms are preserved between the two implementations, available in one of my personal repos [here](https://github.com/turingcompl33t/cpp-playground/tree/master/Algorithm/CustomPredicates). All tests in `test.cpp` pass for both implementations.

krabs/krabs/filtering/comparers.hpp
```diff
@@ -8,9 +8,9 @@
 #pragma warning(disable: 4800) // bool warning in boost when _MANAGED flag is set
 #endif
 
+#include <algorithm>
+
 #include "../compiler_check.hpp"
-#include <boost/algorithm/string.hpp>
```
davemcincork, line 12: Because you've got rid of boost, you can also get rid of the pragma warning stuff above and below.

krabs/krabs/filtering/comparers.hpp
```diff
@@ -64,10 +60,20 @@ namespace krabs { namespace predicates {
             template <typename Iter1, typename Iter2>
             bool operator()(Iter1 begin1, Iter1 end1, Iter2 begin2, Iter2 end2) const
             {
-                auto r1 = boost::make_iterator_range(begin1, end1);
-                auto r2 = boost::make_iterator_range(begin2, end2);
+                const auto dist1 = std::distance(begin1, end1);
```
davemcincork, line 63: This works for me here:
```C++
const auto dist1 = std::distance(begin1, end1);
const auto dist2 = std::distance(begin2, end2);
return dist1 >= dist2 && std::equal(begin1, begin1 + dist2, begin2, Comparer());
```
Note that std::equal() returns true when distance is 0.

krabs/krabs/filtering/comparers.hpp
```diff
@@ -80,10 +86,20 @@ namespace krabs { namespace predicates {
...
```
davemcincork, line 89: As above, this works for me here:
```C++
const auto dist1 = std::distance(begin1, end1);
const auto dist2 = std::distance(begin2, end2);
return dist1 >= dist2 && std::equal(end1 - dist2, end1, begin2, Comparer());
```

krabs/krabs/filtering/comparers.hpp
```diff
@@ -48,10 +45,9 @@ namespace krabs { namespace predicates {
             template <typename Iter1, typename Iter2>
             bool operator()(Iter1 begin1, Iter1 end1, Iter2 begin2, Iter2 end2) const
             {
-                auto r1 = boost::make_iterator_range(begin1, end1);
-                auto r2 = boost::make_iterator_range(begin2, end2);
-
-                return boost::contains(r1, r2, Comparer());
+                // semantics of boost::contains() require that empty range is always contained
+                return std::distance(begin2, end2) == 0 ||
```
davemcincork, line 49: No need for the special handling for when dist2 is 0 because since C++11, the standard requires std::search() to return begin1 instead of end2 in this case.

swannman:
Also adding @kylereedmsft who may be interested in reviewing.

krabs/krabs/filtering/comparers.hpp
kallanreed: Take a look at std::mismatch - it may be possible to use that instead of having to move the iterators around.
I wouldn't rely on iterator math working - that only works with random access iterators.

krabs/krabs/filtering/comparers.hpp
kallanreed: As above, avoid the iterator math to avoid dependency on RandomAccess iterators.

krabs/krabs/filtering/comparers.hpp
turingcompl33t: I appreciate the std::mismatch suggestion - rewrote it in terms of that algorithm and it is much more concise.

krabs/krabs/filtering/comparers.hpp
turingcompl33t: Good point, thanks!

krabs/krabs/filtering/comparers.hpp
turingcompl33t: Agreed. Problem here is when both the target sequence and the test sequence are empty. True to the standard, std::search() still returns begin1, but in this case begin1 == end1, so ultimately the function erroneously returns false. The check against an empty test sequence is the most expressive way I can see to handle this because it is the only case in which an empty target sequence contains any test sequence, so we can short-circuit the std::search() test.

krabs/krabs/filtering/comparers.hpp
```diff
@@ -45,9 +36,9 @@ namespace krabs { namespace predicates {
             template <typename Iter1, typename Iter2>
             bool operator()(Iter1 begin1, Iter1 end1, Iter2 begin2, Iter2 end2) const
             {
-                // semantics of boost::contains() require that empty range is always contained
-                return std::distance(begin2, end2) == 0 ||
-                    std::search(begin1, end1, begin2, end2, Comparer()) != end1;
+                // empty test range always contained, even when input range empty
+                return std::distance(begin2, end2) == 0
```
davemcincork, line 40: Might be marginally more performant to use ``begin2 == end2`` as the first part of this test instead of ``std::distance(begin2, end2) == 0``.

krabs/krabs/filtering/comparers.hpp
davemcincork: I guess I was thinking of these functors being used only on strings and not arbitrary, non-random access containers (in fact I'm not sure why I even used std::distance() instead of just endN-beginN). I take the point that avoiding the assumption of random access makes these as general as possible, and that a good template should be as general as possible. But when it comes to string comparison in particular we don't want a quest for generality to be (a) slower and (b) harder to understand/maintain.

krabs/krabs/filtering/comparers.hpp
turingcompl33t: Can't argue with you there; no reason for the additional function call.

turingcompl33t:
@davemcincork agree with the naming update. It appears that both STL and Boost follow this convention internally, so would be foolhardy to buck that tradition here.

krabs/krabs/filtering/comparers.hpp
kallanreed: Those functions are idiomatic c++ and are optimized away correctly for the concrete instantiation of the template. They are not slower.

swannman:
@turingcompl33t could you also remove the `boost` dependency from the *.vcxproj, packages.config, and krabsetw.nuspec files? We can bump the version in krabsetw.nuspec as well. Thanks!

turingcompl33t:
Got it @swannman. Should the version in O365.Security.Native.ETW.nuspec and O365.Security.Native.ETW.Debug.nuspec be bumped as well? Unsure of what the policy is on keeping the version across the three consistent.

swannman:
Fair point - let's bump the version in all three .nuspec files to maintain consistency.

PR #127, opened 2020-04-22
Fix compiler errors when winternl.h is included before krabs.hpp.

dravenson:
A recent change to krabsetw partially defines an enum that is already defined in winternl.h, and it also declares an NT system call that is already declared in winternl.h. Both of these things cause compiler errors in any project which includes krabs.hpp and winternl.h. This PR adds some preprocessor conditionality to avoid defining and declaring if winternl.h has already been included.


dravenson:
> 
> 
> Thanks for fixing this! Could you bump the version in `krabsetw.nuspec` so projects that consume these headers via nuget can pick up the change?

Sure. There were three .nuspec files in the project root and all had the same version number. So I have incremented all three.

PR #130, opened 2020-05-04
Bring back process api for trace

jrave:
The separate process API allows more fine grained control in multi threaded scenarios as there is no possibility of an implicit opening of the trace.

### DIFF
--- a/O365.Security.Native.ETW.Debug.nuspec
+++ b/O365.Security.Native.ETW.Debug.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Native.ETW.Debug</id>
-        <version>4.1.9</version>
+        <version>4.1.10</version>
         <title>Microsoft.O365.Security.Native.ETW Debug - managed wrappers for krabsetw</title>
         <authors>Microsoft</authors>
         <owners>Microsoft</owners>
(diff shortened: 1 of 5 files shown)


krabs/krabs/etw.hpp
```diff
@@ -62,6 +62,14 @@ namespace krabs { namespace details {
         */
         EVENT_TRACE_LOGFILE open();
 
+		/**
```
swannman, line 65: nit: please replace tabs with spaces (also in `trace.hpp`)

O365.Security.Native.ETW.Debug.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Native.ETW.Debug</id>
-        <version>4.1.9</version>
+        <version>4.1.10</version>
```
swannman, line 5: thanks for updating the nuget version! please also update the `<releaseNotes>` tab in these three `.nuspec` files to describe your change.

jrave:
Updated the PR to address the mentioned issues.

jdu2600:
Sorry for removing this. The use case in the original ticket seemed redundant, and we couldn't think of another one at the time. 

Should we add this to the C++/CLI API also to keep them consistent?

swannman:
I'm okay with not exposing this to C++/CLI unless someone in the future has a use case.

PR #131, opened 2020-05-04
fixed schema event_name parsing

pathtofile:
Fixes two things:
1) `event_name` was using at `OpcodeNameOffset` instead of `EventNameOffset`
2) `EventNameOffset` can be 0, in which case it doesn't have an event name

### DIFF
--- a/krabs/krabs/schema.hpp
+++ b/krabs/krabs/schema.hpp
@@ -248,9 +248,20 @@ namespace krabs {
 
     inline const wchar_t *schema::event_name() const
     {
-        return reinterpret_cast<const wchar_t*>(
-            reinterpret_cast<const char*>(pSchema_) +
-            pSchema_->EventNameOffset);
+        /*
+        EventNameOffset will be 0 if the event does not have an assigned name or
+        if this event is decoded on a system that does not support decoding
+        manifest event names. Event name decoding is supported on Windows
+        10 Fall Creators Update (2017) and later.
+        */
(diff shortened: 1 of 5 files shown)


krabs/krabs/schema.hpp
```diff
@@ -248,9 +248,20 @@ namespace krabs {
...
-        return reinterpret_cast<const wchar_t*>(
-            reinterpret_cast<const char*>(pSchema_) +
-            pSchema_->OpcodeNameOffset);
+        /*
+        EventNameOffset will be 0 if the event does not have an assigned name or
+        if this event is decoded on a system that does not support decoding
+        manifest event names. Event name decoding is supported on Windows
+        10 Fall Creators Update (2017) and later.
+        */
+        if (pSchema_->EventNameOffset != 0) {
```
swannman, line 272: @pathtofile good catch! This change is correct since `event_name` will now return the value at `EventNameOffset1. However, existing callers may have been depending on the value at `OpcodeNameOffset`. Could you add a new `opcode_name` property that returns the value at `OpcodeNameOffset`? Thanks!

pathtofile: No worries, added.
Also, if we're considering people may have been depending on the value at `OpcodeNameOffset` to be available via  `event_name`, it seems like we should bump the version number and add the release notes, to make this change is obvious to people.

So I've also gone and done that :-)

PR #132, opened 2020-05-10
Enable parsing TDH_IN_POINTER types

pathtofile:
This merge attempts to enable easy parsing of `TDH_INTYPE_POINTER`, whilst attempting to protect developers from poor decisions.

Whilst `TDH_INTYPE_POINTER` are Pointers, the likelihood the address is accessible to the trace process is extremely low, and at times might even be pointers to kernel memory.
As such I feel any solution needs to make it clear to developers using Krabs that while it is OK to get the pointer for printing or comparison purposes, any attempt at de-referencing it will probably blow up in their face.

If we just enable `.parse<void*>()` it as a `void*`, then a dev could blindly do:
```c
std::uint32_t* pFileobj = (std::uint32_t*)parser.parse<void*>(L"FileObject");

// This will either explode (esp. if a kernel pointer), or point to something else entirely
std::uint32_t fileobj = *pFileobj ;

// If the above didn't explode, then this could be altering something else
fileobj += 5;
```

That's why my attempt at a solution opted to create a new struct type `pointer`, with the reference named `wild_reference`. This makes the developer have to write:
```c
krabs::pointer krabs_fileobj = parser.parse<krabs::pointer>(L"FileObject");

// Have to use 'wild' reference
// Also const prevents alteration
const std::uint32_t* pFileobj_wild = (const std::uint32_t*)(krabs_fileobj.wild_reference);

// This will still go boom, but we gave them a bit a warning.
const std::uint32_t fileobj = *pFileobj_wild;
```

We could also use `dangerous` instead of `wild` if you feel that would be better.

Another option would make `krabs::pointer.wild_reference` an integer instead of a `void*`, make it even more explicit that the pointer isn't for de-referencing.

### DIFF
--- a/Microsoft.O365.Security.Native.ETW/EventRecord.hpp
+++ b/Microsoft.O365.Security.Native.ETW/EventRecord.hpp
@@ -351,6 +351,53 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         }
 #pragma endregion
 
+#pragma region Pointer
+        /// <summary>
+        /// Get a Pointer type from the from the specified property name
+        /// </summary>
+        /// <param name="name">property name</param>
+        /// <returns>the IntPtr value associated with the specified property</returns>
+        virtual IntPtr^ GetPointer(String^ name)
+        {
+            const auto& addr = GetValue<krabs::pointer>(name);
+            return ConvertToPointer(addr);
+        }
+
+        /// <summary>
+        /// Get a Pointer from the specified property name or returns
+        /// the specified default value.
+        /// </summary>
+        /// <param name="name">property name</param>
+        /// <param name="defaultValue">the default value to return if the property lookup fails</param>
+        /// <returns>the SocketAddress value associated with the specified property or the specified default value</returns>
+        virtual IntPtr^ GetPointer(String^ name, IntPtr^ defaultValue)
+        {
+            IntPtr^ addr;
+
+            if (TryGetPointer(name, addr))
+                return addr;
+
+            return defaultValue;
+        }
+
+        /// <summary>
+        /// Attempt to get a Security Identifier (SID) from the specified property name.
+        /// </summary>
+        /// <param name="name">property name</param>
+        /// <param name="result">the resulting SocketAddress</param>
+        /// <returns>true if fetching the SocketAddress succeeded, false otherwise</returns>
+        virtual bool TryGetPointer(String^ name, [Out] IntPtr^% result)
+        {
+            krabs::pointer addr;
+            bool success = TryGetValue(name, addr);
+
+            if (success)
+                result = ConvertToPointer(addr);
+
+            return success;
+        }
+#pragma endregion
+
 #pragma region DateTime
         /// <summary>
         /// Get a DateTime from the specified property name.
@@ -795,6 +842,12 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         {
             auto managed_string = gcnew String(addr.sid_string.c_str());
             auto managed = gcnew SecurityIdentifier(managed_string);
+            return managed;
+        }
+
+        IntPtr^ ConvertToPointer(const krabs::pointer& addr)
+        {
+            auto managed = gcnew IntPtr(reinterpret_cast<long long>(addr.wild_reference));
             return managed;
         }
 
--- a/Microsoft.O365.Security.Native.ETW/EventRecord.hpp
+++ b/Microsoft.O365.Security.Native.ETW/EventRecord.hpp
@@ -356,7 +356,7 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         /// Get a Pointer type from the from the specified property name
         /// </summary>
         /// <param name="name">property name</param>
-        /// <returns>the IntPtr value associated with the specified property</returns>
+        /// <returns>The IntPtr associated with the specified property</returns>
         virtual IntPtr^ GetPointer(String^ name)
         {
             const auto& addr = GetValue<krabs::pointer>(name);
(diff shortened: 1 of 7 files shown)


Microsoft.O365.Security.Native.ETW/EventRecord.hpp
```diff
@@ -351,6 +351,53 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
+            IntPtr^ addr;
+
+            if (TryGetPointer(name, addr))
+                return addr;
+
+            return defaultValue;
+        }
+
+        /// <summary>
+        /// Attempt to get a Security Identifier (SID) from the specified property name.
```
swannman, line 384: copy/paste error on the comments?

krabs/krabs/parser.hpp
```diff
@@ -367,14 +367,14 @@ namespace krabs {
...
         //      on an 8 - byte boundary, so its size is 8 bytes on a
         //      32 - bit computer and 16 bytes on a 64 - bit computer.
         //      Doubling the pointer size handles both cases.
-        //
-        // So as we are 64-bit, this means we the SID starts at
-        // 16 bytes into the data
-        static const int sid_start = 16;
+        ULONG sid_start = 16;
+        if (EVENT_HEADER_FLAG_32_BIT_HEADER == (schema_.record_.EventHeader.Flags & EVENT_HEADER_FLAG_32_BIT_HEADER)) {
+            sid_start = 8;
```
swannman, line 376: thank you for adding this! I wondered whether we needed special handling for 32-bit here.

krabs/krabs/parse_types.hpp
```diff
@@ -253,6 +253,42 @@ namespace krabs {
...
 
+    /**
+    * <summary>
+    * Used to handle parsing of POINTER types
+    */
+    struct pointer {
+        // Make the reference const and call it wild,
+        // As any attempt to derefernce or alter what it points to
+        // would be very undefined behaviour, as it
+        // is unlikley it is pointing to inside our process' memory
```
swannman, line 264: typo `unlikely`

krabs/krabs/parse_types.hpp
```diff
@@ -253,6 +253,42 @@ namespace krabs {
     private:
     };
 
+    /**
+    * <summary>
+    * Used to handle parsing of POINTER types
+    */
+    struct pointer {
+        // Make the reference const and call it wild,
+        // As any attempt to derefernce or alter what it points to
```
swannman, line 262: typo `dereference`

swannman:
cc @kylereedmsft who may have thoughts on `krabs::pointer.wild_reference`. The current proposal looks reasonable to me but I'm not sure whether there is already an idiomatic way to handle this?

krabs/krabs/parse_types.hpp
```diff
@@ -253,6 +253,42 @@ namespace krabs {
...
+        // would be very undefined behaviour, as it
+        // is unlikley it is pointing to inside our process' memory
+        const void* wild_reference;
```
kallanreed, line 265: Kind of a strange name. Few thoughts: 1) why even store it as an actual pointer at all? It's probably never a good idea to use it as a pointer in code. You could just keep a uint64_t value and let the top bytes be empty on x86. Just call it value or something. Alternately, you could make wrap this in a a type that doesn't allow dereferencing at all (just overload operator* and make it assert, but ultimately you'll still want to see the value at some point, so it seems like using an integer may be better

If you really want a pointer (so things like pointer printf specifiers work), I'd still suggest calling it `value` instead of `wild_reference`.

krabs/krabs/parse_types.hpp
```diff
@@ -253,6 +253,42 @@ namespace krabs {
...
+
+        static pointer from_bytes(const BYTE* bytes, size_t size_in_bytes)
+        {
+            pointer pt;
+
+            // If 32-Bit, first parse as a uint32
+            // Then we can 'cast' that as pointer
+            if (size_in_bytes == sizeof(uint32_t)) {
+                uint64_t tmp = *(uint32_t*)bytes;
```
kallanreed, line 274: I think we prefer static_casts in this code. Just be consistent with the rest of the file

krabs/krabs/parse_types.hpp
jdu2600: Hey Pat,

Nice work.  :-)

I think that using a uint64_t should be enough of a warning to the developer.
And perhaps just 'address' if it needs a name?

John

For reference -
perfview's TraceEvent defines an 'Address' type which is just a UInt64.
And the default string representation is 0x%x.

https://github.com/microsoft/perfview/blob/9d8a9b90c5ed917da1b9bdea98210114a270e633/src/TraceEvent/TraceEvent.cs#L20
https://github.com/microsoft/perfview/blob/9d8a9b90c5ed917da1b9bdea98210114a270e633/src/TraceEvent/TraceEvent.cs#L978

pathtofile:
@jdu2600 @kallanreed @swannman - So I'm happy with instead making it just an `uint64_t`, as I also don't really think people should be attempting to use the pointer. Summary of changes:
* Changed `struct pointer` to parse and hold address as a `uint64_t` instead of a `void*`
* Renamed address variable to "address", to make the name clearer.
* Fixed various typos from bad copy/paste or keyboarding.

pathtofile:
RE: https://github.com/microsoft/krabsetw/pull/132#discussion_r423197793 @kallanreed it had to be a `reintepret_cast` as it doesn't like the `byte*`->`uint32_t*` conversion, but I took your comment to mean 'be consistent with using c++ style casts' and fixed it

swannman:
@kallanreed any review comments before I merge?

kylereedmsft:
Looks ok to me.



________________________________

pathtofile:
@swannman updated the nuspec

Issue #66, opened 2019-01-23
WPP trace message capture


udoe:
Hi,

I'm trying to create an application that captures events produced by WPP in a kernel-mode driver and I found krabsetw quite useful for this task. Basically creating and starting a session using krabs::user_trace works as expected. 

The issue is that ut::forward_events() checks the record.EventHeader.ProviderId in each event and forwards it only if this field matches the provider GUID. I found that in WPP messages the EventHeader.ProviderId field is not set to the provider GUID. It is set to some local GUID generated by WPP instead. So WPP messages cannot pass the check in ut::forward_events().

The [documentation](https://docs.microsoft.com/en-us/windows/desktop/api/tdh/nf-tdh-tdhgetproperty) states that the EVENT_HEADER_FLAG_TRACE_MESSAGE flag indicates a WPP message. So I fixed that by modifying the check as follows:

 if (0 != (record.EventHeader.Flags & EVENT_HEADER_FLAG_TRACE_MESSAGE) || 
          record.EventHeader.ProviderId == provider.get().guid_) {
      provider.get().on_event(record);
  }

It would be great if this check could be made configurable. 
Or could it be removed completely?
Maybe there should be a separate trace type for WPP, e.g. krabs::wpp_trace.

Thanks
Udo

zacbrown:
Hi Udo,

I haven't personally used WPP with krabsetw so I'm not surprised there's a subtle issue there. We'd happily accept a patch for that if you wanted to create one.

thanks,

Zac

pathtofile:
So FWIW I just tried to use the latest Krabs with the AMSI WPP trace in Matt Graeber's blog (https://posts.specterops.io/data-source-analysis-and-dynamic-windows-re-using-wpp-and-tracelogging-e465f8b653f7)

And it seemed to work ok? The event_names weren't parsed correctly, but I did recieve the events and could parse the data in them.

Tried some other WPP traces and the same thing - Event Names were wrong, but data still there

swannman:
Thanks @pathtofile! We are discussing the fix on your PR #131.

swannman:
Resolved by #131

PR #135, opened 2020-08-01
added a WPP trace example (#66)

jdu2600:
Chatted to @pathtofile about WPP tracing in krabs and got some hints on how to make this work.

Thought that folks might appreciate an example - since it's a little different to other provider types.

### DIFF
--- a/examples/ManagedExamples/Program.cs
+++ b/examples/ManagedExamples/Program.cs
@@ -15,6 +15,7 @@ namespace ManagedExamples
             //UserTrace004.Start();
             //UserTrace005.Start();
             //FakingEvents001.Start();
+            //WppTrace001.Start();
         }
     }
 }
--- a/examples/ManagedExamples/WppTrace001.cs
+++ b/examples/ManagedExamples/WppTrace001.cs
@@ -1,0 +1,128 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// This example shows how to use a UserTrace to monitor WPP providers.
+// This is a special case due slight differences in the event format,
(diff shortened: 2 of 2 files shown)


pathtofile:
Awesome work bud, definetly a useful example!
The message GUIDs were different on my machine, possibly something that changes with dll versions?

This raises a bit of a pain point with WPP tracing with Krabs - You subscribe to 1 provider GUID, but messages come on another. If you don't know this beforehand, then the messages get silently dropped in `ut::forward_events`.

I wonder if there's a way we could improve that, maybe by enabling a callback function for any provider in the session, or an 'unhandled provider' callback...

jdu2600:
Different message guids? So my example didn't work for you? (You saw that I wasn't using Matt's amsi WPP example, right?)

In #66 Udo suggested checking the EVENT_HEADER_FLAG_TRACE_MESSAGE flag - and always forwarding WPP events.
But what if you had subscribed to two or more WPP providers? Each one would receive every event from all WPP providers...so you'd still need to filter on the message GUID in order to find the right events. (iirc Event Id is insufficient as a filter even for a single WPP provider. You need both message GUID and Event Id to know the appropriate UserData format).

The current approach in my example (while perhaps not semantically pure - i.e. using a message GUID as a provider GUID) achieves the necessary level of filtering. 

The missing piece imho is helper functions for manually parsing the WPP UserData. (These actually probably already exist in krabs - and might just need to be exposed?) But I'll leave that for someone who is making heavy use of WPP providers to do. :-)
Maybe the author of sealighter might do it if they decide to extend their WPP support? ;-)

That said, I recall that Microsoft's EventTrace library allowed you to register a callback for unhandled events on a provider. And this was quite a useful feature.  :-)

swannman:
Thanks for adding this!

PR #140, opened 2020-11-06
Ignore struct padding when computing std::hash<schema_key> 

### LINKED ISSUE #139: schema_key::operator== is inconsistent with std::hash<schema_key> specialisation
This issue manifests in my application as a memory leak in the schema_locator:
- The ``schema_key::operator==`` overload does a member-wise comparison of all five fields in the struct.
- The ``std::hash<schema_key>`` specialisation acts upon all bytes in the struct, including the padding bytes that are not evaluated by ``schema_key::operator==``.
- The result is that two keys which compare equal will not necessarily hash equal.
- This is a problem because for std::unordered_map to work correctly, it is necessary that ``key1 == key2`` => ``hash(key1) == hash(key2)``.
- And this causes the schema_locator to have duplicate entries for the same event ID.

The following test program illustrates the problem:
```C++
#define WIN32_LEAN_AND_MEAN
#include <krabs.hpp>

int main()
{
    const EVENT_RECORD eventRecord; // uninitialised garbage but that's OK

    const krabs::schema_key key1{ eventRecord };
    const krabs::schema_key key2{ eventRecord };

    const std::hash<krabs::schema_key> hash;

    auto exitCode = EXIT_SUCCESS;

    if ((key1 == key2) != (hash(key1) == hash(key2)))
    {
        fprintf(stderr, "Oops!\n");
        exitCode = EXIT_FAILURE;
    }

    return exitCode;
}
```
A Release build done with VS2019.latest will output "Oops!". A Debug build will not because the stack is initialised.

swannman:
Fixes #139

### DIFF
--- a/krabs/krabs/guid.hpp
+++ b/krabs/krabs/guid.hpp
@@ -309,4 +309,21 @@ namespace std
 
         return { managed.get() };
     }
+
+    template<>
+    struct std::hash<krabs::guid>
+    {
+        size_t operator()(const krabs::guid& guid) const
+        {
+            const char* guidBytes = (const char*)&guid;
+            size_t h = 2166136261;
+
+            for (auto i = 0; i < sizeof(guid); ++i)
+            {
+                h ^= (h << 5) + (h >> 2) + guidBytes[i];
+            }
+
+            return h;
+        }
+    };
 }
--- a/krabs/krabs/guid.hpp
+++ b/krabs/krabs/guid.hpp
@@ -315,9 +315,10 @@ namespace std
     {
         size_t operator()(const krabs::guid& guid) const
         {
-            const char* guidBytes = (const char*)&guid;
+            // Shift-Add-XOR hash starts with this prime number
             size_t h = 2166136261;
 
+            const char* guidBytes = (const char*)&guid;
             for (auto i = 0; i < sizeof(guid); ++i)
             {
                 h ^= (h << 5) + (h >> 2) + guidBytes[i];
--- a/krabs/krabs/schema_locator.hpp
+++ b/krabs/krabs/schema_locator.hpp
@@ -29,7 +29,6 @@ namespace krabs {
      * Type used as the key for cache lookup in a schema_locator.
      * </summary>
      */
-    #pragma pack(push, 1) // Ref: https://github.com/microsoft/krabsetw/issues/139
     struct schema_key
     {
         guid      provider;
@@ -56,7 +55,6 @@ namespace krabs {
 
         bool operator!=(const schema_key &rhs) const { return !(*this == rhs); }
     };
-    #pragma pack(pop)
 }
 
 namespace std {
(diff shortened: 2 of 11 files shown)


henriblMSFT (approved):

zacbrown (dismissed):

krabs/krabs/guid.hpp
```diff
@@ -309,4 +309,21 @@ namespace std
...
         return { managed.get() };
     }
+
+    template<>
+    struct std::hash<krabs::guid>
+    {
+        size_t operator()(const krabs::guid& guid) const
+        {
+            const char* guidBytes = (const char*)&guid;
+            size_t h = 2166136261;
```
zacbrown, line 319: Observation - might be worthwhile to indicate where this magic number came from.

krabs/krabs/schema_locator.hpp
```diff
@@ -70,13 +70,13 @@ namespace std {
...
         {
             // Shift-Add-XOR hash - good enough for the small sets we deal with
-            const char* p = (const char*)&key;
             size_t h = 2166136261;
 
-            for(auto i = 0; i < sizeof(key); ++i)
-            {
-                h ^= (h << 5) + (h >> 2) + p[i];
-            }
+            h ^= (h << 5) + (h >> 2) + std::hash<krabs::guid>()(key.provider);
```
swannman, line 75: should we instantiate std::hash<krabs::guid>() once instead of each time here?

swannman: should `krabs::guid` expose a `hash` member that lazily computes its `std::hash` the first time it's requested?

tests/krabstests/test_schema_key.cpp
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include "CppUnitTest.h"
+#include <krabs.hpp>
+
+using namespace Microsoft::VisualStudio::CppUnitTestFramework;
+
+namespace krabstests
```
kylereedmsft, line 9: Did you try these tests with a release build just to make sure the OP bug is covered?

kylereedmsft (approved):

krabs/krabs/schema_locator.hpp
kylereedmsft: Just for future ref: the hasher instance has no initialization overhead. The guid instances are short lived and I _think_ the hash is only computed once.

tests/krabstests/test_schema_key.cpp
swannman: I did! :)

krabs/krabs/guid.hpp
```diff
@@ -309,4 +309,21 @@ namespace std
 
         return { managed.get() };
     }
+
+    template<>
+    struct std::hash<krabs::guid>
+    {
+        size_t operator()(const krabs::guid& guid) const
```
davemcincork, line 316: I know this PR is already merged but FWIW...

I find it strange to calculate a byte-by-byte hash on something that is already highly random. A good specialisation of std::hash just has to have low risk of collision. So if you just take the lower sizeof(size_t) bits of the 128-bit GUID you’ve already got a very unique value, i.e. just return:

reinterpret_cast<const size_t*>(&guid)[(sizeof(guid) / sizeof(size_t)) - 1]

kylereedmsft: That's a good point.
I was taking a look at the .NET guid hash which at least treats it as blocks and uses the whole value. May be worthwhile.
There are different guid versions with varying amounts of randomness in the different parts - but given that these are all provider guids, I'd guess they are all the random kind.
```
        public override int GetHashCode()
        {
            return _a ^ (((int)_b << 16) | (int)(ushort)_c) ^ (((int)_f << 24) | _k);
        }```

swannman: Good catch, I applied the recommended improvement in #144

Issue #143, opened 2020-11-12
".NET Common Language Runtime" events not captured


jdu2600:
I've recently discovered that {e13c0d23-ccbc-4e12-931b-d9cc2eee27e4} is a very interesting provider.

In .NET 4.8 it is a modern manifest provider called ['Microsoft-Windows-DotNETRuntime'](https://gist.github.com/jdu2600/c6ecb2ed684c6cac69dfe1d32b2982c8)
In .NET 3.5 it is a legacy MOF provider called ['.NET Common Language Runtime'](https://gist.github.com/jdu2600/e99c95d87ac20a44c47068b71caf333b)

Currently krabs receives, but does not correctly forward, the events generated by the MOF provider.
I can workaround the forwarding - but krabs still does not seem to correctly parse the event.
Note - I haven't investigated yet whether krabs doesn't support all MOF providers - or just this special cae.

Microsoft's [TraceEvent](https://github.com/microsoft/perfview/tree/master/src/TraceEvent) library correctly filters and parses events from both sources - and could be an exemplar to determine any necessary steps. 
e.g. `SilkETW.exe -t user -pn Microsoft-Windows-DotNETRuntime -uk 0x8 -ot file -p C:\\Some\\Path\\out.json` will collect loader events from both .NET runtimes.

jdu2600:
Looks like krabs didn't support all MOF providers at all.

Microsoft's [TDH documentation](https://docs.microsoft.com/en-us/windows/win32/etw/retrieving-event-data-using-tdh) nicely describes how to handle both cases.

swannman:
Resolved by #145

fs-granma:
I also was hit by this issue. The events do get forwarded now but they hit "Could not find the schema" exception. I suppose it is related to "Looks like krabs didn't support all MOF providers at all.".

What would it take to support MOF providers?

jdu2600:
Hmm. I didn't test on Windows 7.
To add the MOF provider support I actually had to retrieve the schema in order to determine which provider to forward the event to...so the schema should definitely be available...

Which version of krabs are you using?

fs-granma:
The krabs version I'm using is 4.1.16.

Something doesn't quite add up.

logman output would indicate that the MOF provider is present:
```
Provider                                 GUID
-------------------------------------------------------------------------------
.NET Common Language Runtime             {E13C0D23-CCBC-4E12-931B-D9CC2EEE27E4}
```

[...]

PR #145, opened 2020-11-14
add support for MOF providers, and update WPP example (#143)

jdu2600:


krabs/krabs/ut.hpp
```diff
@@ -194,11 +194,33 @@ namespace krabs { namespace details {
         const EVENT_RECORD &record,
         const krabs::trace<krabs::details::ut> &trace)
     {
-        for (auto &provider : trace.providers_) {
+        auto found = false;
+
+        // skip event trace header events
+        if (record.EventHeader.ProviderId == EventTraceGuid) {
```
swannman, line 200: under what conditions will this be true? should we only do this if we weren't able to find a manifest provider below?

swannman (changes requested):
Thanks for submitting this! I reviewed and added a few comments/questions.

examples/ManagedExamples/WppTrace001.cs
```diff
@@ -70,59 +67,78 @@ public static void Start()
...
-            
+
             // Side note - if you want to turn up the verbosity of your COM WPP diagnostic tracing, then enable
             // OLE32 tracing via the registry following the instruction here -
             // https://support.microsoft.com/en-us/help/926098/how-to-enable-com-and-com-diagnostic-tracing
             //
             // Alternatively call _ControlTracing (4) via combase's 18f70770-8e64-11cf-9af1-0020af6e72f4 RPC interface.
 
+            trace.Enable(allWppDummyProvider);
             trace.Enable(ole32WppProvider);
```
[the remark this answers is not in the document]
jdu2600: It's the same example.

The ole32WppProvider enables the events, but the allWppDummyProvider receives the events.

krabs/krabs/ut.hpp
jdu2600: EventTraceGuid is the 'log header event'.
https://docs.microsoft.com/en-us/windows/win32/etw/eventtraceevent

All of the MSDN examples skip this event. I followed suit - but really mostly so that I could skip these events while debugging and hit the first 'real' event sooner.

I also had a vague idea that this would make it more performant. i.e. skipping unnecessary comparisons in the loops.
But on reflection, as these are header-only events, it makes more sense overall to not include this check.

jdu2600: Plus technically someone could subscribe to the header events.

examples/ManagedExamples/WppTrace001.cs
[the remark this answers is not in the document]
jdu2600: I updated the comments to make this clearer.

krabs/krabs/ut.hpp
```diff
@@ -194,11 +194,28 @@ namespace krabs { namespace details {
...
             if (record.EventHeader.ProviderId == provider.get().guid_) {
+                found = true;
                 provider.get().on_event(record, trace.context_);
             }
         }
+
+        // for MOF providers, EventHeader.Provider is the *Message* GUID
+        // we need to ask TDH for event information in order to determine the
+        // correct provider to pass this event to
+        if (!found) {
```
swannman, line 210: nit: to reduce nesting, consider doing this:

```
if (found) return;

// for MOF providers, EventHeader.Provider is the *Message* GUID
// we need to ask TDH for event information in order to determine the
// correct provider to pass this event to
auto schema = get_event_schema_from_tdh(record);
....
```

swannman (changes requested):
This looks great. I added one more nitpick suggestion, and then I'll merge this in!

Thanks for clarifying the example code, I didn't completely follow it the first time through. Clever solution.

krabs/krabs/ut.hpp
jdu2600: We could even eliminate the found variable entirely - and just return directly from within the manifest provider loop?

This would mean that only a single provider could be registered for a given GUID. That seems like a fairly sane assumption to make?
In fact, I suspect that this assumption is already implicit with ETW. i.e. a session can't have two sets of differing flags/levels/native filters for the same GUID.

swannman (approved):

jdu2600:
^^^^ @pathtofile - MOF provider support

PR #149, opened 2020-12-07
Add compatibility with 'phnt' library (https://github.com/processhacker/phnt)

starix:


swannman:
@starix thanks for submitting this! We discussed internally and, rather than adding `#ifndef` for specific external libraries like Process Hacker, we would prefer an approach where Krabs wraps these definitions in something like `#ifndef KRABS_USE_EXTERNAL_PERFINFO`. Customer using e.g. Process Hacker would `#define KRABS_USE_EXTERNAL_PERFINFO` to opt-in to the new behavior.

Would you be willing to take that approach here?

swannman (changes requested):
@starix thanks for submitting this! We discussed internally and, rather than adding `#ifndef` for specific external libraries like Process Hacker, we would prefer an approach where Krabs wraps these definitions in something like `#ifndef KRABS_USE_EXTERNAL_PERFINFO`. Customer using e.g. Process Hacker would `#define KRABS_USE_EXTERNAL_PERFINFO` to opt-in to the new behavior.

Would you be willing to take that approach here?

starix:
Sorry for delay.  I think last changes do not break library's logic and meet my goals. What do you think about it?

swannman:
@starix will you be consuming these changes in the KrabsETW nuget? If so, could you update the version number [here](https://github.com/microsoft/krabsetw/blob/master/krabsetw.nuspec)?

starix:
@swannman Do I need to fix O365.Security.Native.ETW.nuspec and O365.Security.Native.ETW.Debug.nuspec?

swannman:
@starix if you aren't consuming this via nuget then I don't think we need to update the .nuspec files for this change. Thanks!

swannman (approved):

PR #154, opened 2021-01-10
Save size of data in socket_address struct

starix:


swannman:
@starix thanks for submitting this! Could you tell us a bit about the problem that this solves or the scenario where this was needed?

starix:
@swannman WSAAddressToString function need size as second parameter.

swannman (approved):

PR #155, opened 2021-01-13
Rundown event support and examples

jdu2600:


krabs/krabs/ut.hpp
```diff
@@ -186,7 +197,28 @@ namespace krabs { namespace details {
...
+    }
+
+    inline void ut::enable_rundown(
+        const krabs::trace<krabs::details::ut>& trace)
+    {
+        if (trace.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE)
+            return;
+
+        for (auto& provider : trace.providers_) {
+            if (provider.get().rundown_enabled_) {
```
swannman, line 211: nit: invert the logic and `continue` if `!rundown_enabled_` to reduce nesting

krabs/krabs/kernel_providers.hpp
```diff
@@ -54,43 +54,35 @@ namespace krabs { namespace kernel {
         krabs::guids::debug);
 
     /**
-     * <summary>A provider that enables file I/O name events.</summary>
+     * <summary>A provider that enables basic disk I/O events.</summary>
```
swannman, line 57: Could these changes impact back-compat in the sense that callers using `CREATE_CONVENIENCE_KERNEL_PROVIDER(disk_io_provider)` may now see unexpected events, or callers using `CREATE_CONVENIENCE_KERNEL_PROVIDER(disk_file_io_provider)` will now get a compilation error since the symbol is no longer defined?

I wonder whether changes in this file should always be additive (add a new convenience provider in addition to previous ones) or purely a rename (name changes but flags stay the same).

swannman (changes requested):
@jdu2600 thanks for submitting this -- this is super-clean! I added a couple of comments to consider.

krabs/krabs/kernel_providers.hpp
zacbrown: I tend to agree with @swannman.

At a minimum, even if it doesn't break anything because the new events would be ignored, you're still adding more events to process on downstream consumers.

jdu2600: Sorry. I had meant to flag these changes for discussion. :-/

There is/was a bug (that I originally introduced) in the disk_file_io_provider. The EVENT_TRACE_FLAG_DISK_FILE_IO generates events for the FileIo provider - not the DiskIo provider.

Further, I had originally also enabled EVENT_TRACE_FLAG_DISK_IO based off the documentation here -
 * https://docs.microsoft.com/en-us/windows/win32/api/evntrace/ns-evntrace-event_trace_properties
 * "EVENT_TRACE_FLAG_DISK_FILE_IO0x00000200 | Enables the following FileIo event type (you must also enable EVENT_TRACE_FLAG_DISK_IO)"

But I'm now not convinced this is required... That flag enables *DiskIo* events as documented here -
 * https://docs.microsoft.com/en-us/windows/win32/etw/diskio

That flag is is also not mentioned in the FileIo documentation here -
 * https://docs.microsoft.com/en-us/windows/win32/etw/fileio

So I suggest dropping the EVENT_TRACE_FLAG_DISK_IO flag from the convenience (FileIo) disk_file_io_provider when I fix the bug. It enables DiskIo events which are likely unhandled.


The other changes were a suggested alternate approach to convenience providers with multiple EnableFlags.
For example, there are 3 flags for both the DiskIo and FileIo providers. The documentation implies that one is the primary flag, and the others are optional additional flags. This means that there are up to seven combinations of flags possible - and therefore up to seven possible convenience providers. I was suggesting having at most two convenience providers per GUID - one for the "primary" flag only, and one for all possible flags. 

The current approach is to have one convenience provider per flag. This means that if I use all three convenience providers for a single GUID (e.g. disk_io_provider, disk_init_io_provider and driver_provider) then each event will be delivered three times since krabs matches on the GUID only. Of course, I can create a custom provider with the actual combination of flags required.

I'll revert these changes. The additional convenience is questionable, and not worth breaking backwards compatibility.

swannman (approved):

Issue #157, opened 2021-05-11
Microsoft-Windows-Security-Auditing not work on some versions


zeze-zeze:
I tested the Microsoft-Windows-Security-Auditing in user_trace_005.cpp, which could work in Windows10 19042.928 (host), while it aborted when using Windows Server 2012, Windows 7, Windows10 1079 (vm using vmware). 

In all of the tests I used system privilege to run the program.

Any clue would be grateful !

zeze-zeze:
It turned out that the reason that cause the error was because the error code returned from ControlTrace().
According to [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-controltracea), we can find what error code means.
In Windows10 19042.928, it returns 5, which corresponds to ERROR_ACCESS_DENIED.
In Windows Server 2012 and Windows10 1079, it returns 87, which corresponds to ERROR_INVALID_PARAMETER.

However, I still don't know why the error codes returned are different.

swannman:
Resolved by #158 - thanks @zeze-zeze !

PR #158, opened 2021-05-18
catch invalid_parameter in register_trace()

zeze-zeze:
As [issue #157](https://github.com/microsoft/krabsetw/issues/157) said, there is an error when manage the provider on Microsoft-Windows-Security-Auditing, tested on Windows10 1709, Windows server 2012 R2.

So I handle error code invalid_parameter (87) just the same as the catch for need_to_be_admin_failure (5).

### DIFF
--- a/krabs/krabs/etw.hpp
+++ b/krabs/krabs/etw.hpp
@@ -305,6 +305,12 @@ namespace krabs { namespace details {
                 // StartTrace() actually sets this to 0 on failure
                 trace_.registrationHandle_ = INVALID_PROCESSTRACE_HANDLE;
             }
+            catch (invalid_parameter) {
+                (void)open_trace();
+                close_trace();
+                status = ERROR_SUCCESS;
+                trace_.registrationHandle_ = INVALID_PROCESSTRACE_HANDLE;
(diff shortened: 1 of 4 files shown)


krabs/krabs/etw.hpp
```diff
@@ -305,6 +305,12 @@ namespace krabs { namespace details {
                 // StartTrace() actually sets this to 0 on failure
                 trace_.registrationHandle_ = INVALID_PROCESSTRACE_HANDLE;
             }
+            catch (invalid_parameter) {
+                (void)open_trace();
```
swannman, line 309: could you add a comment on the line after `catch` describing the scenario that you are addressing with this change, so it's clear to future maintainers of this code?

swannman (changes requested):
Thanks for submitting this! I added one minor comment to your change. Could you also bump the minor version number in the three .nuspec files at the project root so we can publish the updated nugets after merging your change?

zeze-zeze:
Done. I have commented on the line after catch and changed the version of .nuspec files.

swannman (approved):

PR #162, opened 2021-06-10
Replace "min" by "std::min<size_t>" to support NOMINMAX

pierricgimmig:
Including krabsetw from a project that defines "NOMINMAX" would cause a
compilation error due to the absence of the "min" macro. The issue is
fixed by using "std::min<size_t>" instead  of "min". Note that we
explicitly specify the template type in "std::min<size_t>" to avoid a
clash with the "min" macro in the case where "NOMINMAX" is not defined.

Also, check that there is no truncation when casting size_t to USHORT.

Tests: ran all test.

### DIFF
--- a/O365.Security.Native.ETW.Debug.nuspec
+++ b/O365.Security.Native.ETW.Debug.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Native.ETW.Debug</id>
-        <version>4.1.19</version>
+        <version>4.1.20</version>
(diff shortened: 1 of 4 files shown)


swannman (approved):
Thanks for doing this! If you are importing krabs via the nuget then please update the version numbers in the three `.nuspec` files at the root of the repo. Otherwise you are good to go!

swannman (approved):

swannman:
@pierricgimmig thanks for updating the `.nuspec` files! Looks like signing the [CLA](https://cla.opensource.microsoft.com/microsoft/krabsetw?pullRequest=162) is the last step and then we can merge this.

pierricgimmig:
Ah yes of course, done. Thanks!

PR #171, opened 2021-09-02
Add optional trace::create() method

pierricgimmig:
Some providers are configured through the use of "TraceSetInformation"
which takes in a session handle. The session handle is available after
the call to "StartTrace" in "register_trace()". Add a "create()" method that
calls "register_trace" independently from "start()" to give an
opportunity to use "TraceSetInformation" with the returned handle
before the blocking call to "start()". Note that "create()" is optional.
If not called, the trace session will be created automatically when
calling "start()", as before.

Note that we could have used "ControlTrace" to obtain the handle of an
existing trace. However, since the session handle was assigned through
the "start()" method, which runs in its own thread, there is a race
condition.

Tests: all tests pass.

### DIFF
--- a/krabs/krabs/etw.hpp
+++ b/krabs/krabs/etw.hpp
@@ -43,6 +43,13 @@ namespace krabs { namespace details {
 
         /**
          * <summary>
+         * Registers the ETW trace identified by the info in the trace type.
+         * </summary>
+         */
+        TRACEHANDLE register_trace();
+
+        /**
+         * <summary>
          * Starts the ETW trace identified by the info in the trace type.
          * </summary>
          */
@@ -95,7 +102,6 @@ namespace krabs { namespace details {
         trace_info fill_trace_info();
         EVENT_TRACE_LOGFILE fill_logfile();
         void close_trace();
-        void register_trace();
         EVENT_TRACE_PROPERTIES query_trace();
         void stop_trace();
         EVENT_TRACE_LOGFILE open_trace();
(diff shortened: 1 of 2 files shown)


pierricgimmig:
Ping. Any thoughts on this? Do you see a better way to retrieve the session handle? To give more context, I need the trace handle in order to initialize stack tracing with `TraceSetInformation`:

```
...
size_t data_size = ids_.size() * sizeof(STACK_TRACING_EVENT_ID);
  ULONG error = TraceSetInformation(trace_handle_, TraceStackTracingInfo, ids_.data(),
                                    static_cast<ULONG>(data_size));
```

I also need to call `TraceSetInformation` with the session handle to set the sampling frequency: 

```
...
ULONG error = TraceSetInformation(trace_handle, TraceSampledProfileIntervalInfo, (void*)&interval,
                                    sizeof(TRACE_PROFILE_INTERVAL));
```

swannman (changes requested):
@pierricgimmig thanks for submitting this! I agree with the outcome - users of the Krabs library should be able to configure trace settings like stack tracing. However, I don't want to expose the internal `TRACEHANDLE` to user code.

Instead, consider adding a `set_trace_information(TRACE_INFO_CLASS InformationClass, PVOID TraceInformation, ULONG InformationLength)` method to `krabs::trace` that operates on the internal trace handle. That preserves encapsulation while allowing our users to configure the trace session.

Would that work?

pierricgimmig:
@swannman , sounds good, thanks, it works indeed. Closing this PR in favor of #181

PR #172, opened 2021-09-02
Add additional filtering to prevent unexpected event IDs when using provider-based filtering

swannman:
In some cases we receive events that should have been filtered by the provider-based filter (EnableFilterDesc) and whose schema cannot be found, leading to a could_not_find_schema exception.

This seems to occur when another trace session is registered on the same machine.

To avoid this, check against the list of provider-based filter event IDs and do not forward the event if it is not on the list.

### DIFF
--- a/O365.Security.Native.ETW.Debug.nuspec
+++ b/O365.Security.Native.ETW.Debug.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Native.ETW.Debug</id>
-        <version>4.1.24</version>
+        <version>4.1.25</version>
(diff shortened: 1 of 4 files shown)


swannman:
This is no longer needed. The source of this issue is another process which improperly opened a manifest-based ETW provider as a .NET EventSource provider, in which case the listener can send commands back to the provider by _creating events on the same provider GUID_ which is wholly unexpected.

Ref section 5.5 [in this EventSource design doc](https://web.archive.org/web/20190112075007/https://msdnshared.blob.core.windows.net/media/MSDNBlogsFS/prod.evol.blogs.msdn.com/CommunityServer.Components.PostAttachments/00/10/32/81/89/StronglyTypedEvents.docx) and `EventCommand` in [EventSource.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L2587).

PR #175, opened 2021-09-04
Send schema errors to OnError callback in all cases

swannman:
A schema lookup occurs when a predicate filter is applied to a record. If the schema lookup fails, a `krabs::could_not_find_schema` exception is thrown. This exception was previously unhandled and caused an exception to propagate on the main trace thread.

A better way to handle this is to catch the schema exception and notify the application via an `OnError` delegate. An `OnError` delegate existed on the C++/CLI `Provider` object but not on the native `event_filter` and `provider` objects nor on the C++/CLI `EventFilter`.

This change finishes the implementation of error handling on the managed and native layers:
- it adds an `OnError` delegate to the C++/CLI `EventFilter` class, wires it up to a new callback in `event_filter`, and adds an appropriate `try`/`catch` in `event_filter::on_event` to use the new callback
- it wires up the existing `OnError` delegate on the C++/CLI `Provider` to a new callback in `provider` and adds an appropriate `try`/`catch` in `base_provider::on_event` to use the new callback
- it adds an interface for EventRecordError to allow OnError handlers to be unit-tested

### DIFF
--- a/Microsoft.O365.Security.Native.ETW/Filtering/EventFilter.hpp
+++ b/Microsoft.O365.Security.Native.ETW/Filtering/EventFilter.hpp
@@ -97,14 +97,20 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         }
 
         void EventNotification(const EVENT_RECORD &, const krabs::trace_context &);
+        void ErrorNotification(const EVENT_RECORD&, const std::string&);
 
     internal:
-        delegate void NativeHookDelegate(const EVENT_RECORD &, const krabs::trace_context &);
+        delegate void OnEventNativeHookDelegate(const EVENT_RECORD &, const krabs::trace_context &);
+        delegate void OnErrorNativeHookDelegate(const EVENT_RECORD&, const std::string&);
 
-        NativeHookDelegate ^del_;
         NativePtr<krabs::event_filter> filter_;
-        GCHandle delegateHookHandle_;
-        GCHandle delegateHandle_;
+        OnEventNativeHookDelegate ^onEventDelegate_;
+        OnErrorNativeHookDelegate ^onErrorDelegate_;
+        GCHandle onEventDelegateHookHandle_;
+        GCHandle onErrorDelegateHookHandle_;
+        GCHandle onEventDelegateHandle_;
+        GCHandle onErrorDelegateHandle_;
+        void RegisterCallbacks();
     };
 
     // Implementation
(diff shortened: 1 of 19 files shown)


O365.Security.Native.ETW.Debug.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Native.ETW.Debug</id>
-        <version>4.1.24</version>
+        <version>4.1.25</version>
```
henriblMSFT, line 5: Should we consider a major version bump or at least a minor version bump? This change the behavior from throwing an exception on the trace to only calling the callback when this type of error occurs. Someone could update krabs and now fail silently. We should also clearly indicate the change on the release note and explain what the consumer needs to change when updating to this version.

Microsoft.O365.Security.Native.ETW/Filtering/EventFilter.hpp
```diff
@@ -97,14 +98,20 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         }
 
         void EventNotification(const EVENT_RECORD &, const krabs::trace_context &);
+        void ErrorNotification(const EVENT_RECORD&, const std::string&);
 
     internal:
-        delegate void NativeHookDelegate(const EVENT_RECORD &, const krabs::trace_context &);
+        delegate void OnEventNativeHookDelegate(const EVENT_RECORD &, const krabs::trace_context &);
```
henriblMSFT, line 104: NIT: This is likely not the correct naming for the delegate. The On is generally reserved for the callback not for the delegate definition or the event definition.

The recommended suffix for the delegate is `EventHandler` but I think it's ok to keep it as Delegate for historical consistency but we should name them
`EventReceivedNativeHookDelegate`
`ErrorReceivedNativeHookDelegate`

See https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-events

krabs/krabs/filtering/event_filter.hpp
```diff
@@ -22,8 +22,10 @@ namespace krabs { namespace details {
 
 namespace krabs {
 
-    typedef void(*c_provider_callback)(const EVENT_RECORD &, const krabs::trace_context &);
```
henriblMSFT, line 25: Is this a breaking change? Would it make more sense if we kept c_provider_callback and just added the c_provider_error_callback? 

We can make the change but that would be another argument for a major version bump

henriblMSFT (commented):
This looks good. Only two things two discuss
A) Let's consider a major version change
B) Let's discuss callback naming conventions. I don't krabs naming is correct right now so we should strive to make it consistent with what krabs does but it was a bit hard to understand what is the convention in Krabs

O365.Security.Native.ETW.Debug.nuspec
swannman: Good call! I bumped the version from 4.1.24 to 4.2.0 and improved the release note.

Microsoft.O365.Security.Native.ETW/Filtering/EventFilter.hpp
swannman: Good suggestion, updated!

krabs/krabs/filtering/event_filter.hpp
swannman: Good point, I made this update to improve readability but we should prioritize back-compat. Reverted.

Microsoft.O365.Security.Native.ETW/Filtering/EventFilter.hpp
swannman: Good call, updated!

swannman:
Thanks @henriblMSFT for the review! Updated with your feedback.

henriblMSFT (approved):

PR #181, opened 2021-09-22
Add trace::set_trace_information method

pierricgimmig:
Some providers need to be configured through ETW's TraceSetInformation
function which we wrap inside trace::set_trace_information to maintain
encapsulation of the trace handle.

Tests: Setup sampling profiling. All tests pass.

See discussion in #171

### DIFF
--- a/krabs/krabs/etw.hpp
+++ b/krabs/krabs/etw.hpp
@@ -179,9 +179,7 @@ namespace krabs { namespace details {
     template <typename T>
     EVENT_TRACE_LOGFILE trace_manager<T>::open()
     {
-        if (trace_.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE) {
-            (void)register_trace();
-        }
+        register_trace();
         enable_providers();
         return open_trace();
     }
@@ -204,10 +202,6 @@ namespace krabs { namespace details {
         PVOID trace_information,
         ULONG information_length)
     {
-        if (trace_.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE) {
-            (void)register_trace();
-        }
-
         ULONG status = TraceSetInformation(
             trace_.registrationHandle_, 
             information_class,
--- a/krabs/krabs/etw.hpp
+++ b/krabs/krabs/etw.hpp
@@ -80,12 +80,13 @@ namespace krabs { namespace details {
         /**
          * <summary>
          * Configures the ETW trace session settings.
+         * See https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-tracesetinformation.
          * </summary>
          */
         void set_trace_information(
-            TRACE_INFO_CLASS informationClass,
-            PVOID traceInformation,
-            ULONG informationLength);
+            TRACE_INFO_CLASS information_class,
+            PVOID trace_information,
+            ULONG information_length);
 
         /**
          * <summary>
@@ -199,9 +200,9 @@ namespace krabs { namespace details {
 
     template <typename T>
     void trace_manager<T>::set_trace_information(
-        TRACE_INFO_CLASS informationClass,
-        PVOID traceInformation,
-        ULONG informationLength)
+        TRACE_INFO_CLASS information_class,
+        PVOID trace_information,
+        ULONG information_length)
     {
         if (trace_.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE) {
             (void)register_trace();
@@ -209,9 +210,9 @@ namespace krabs { namespace details {
 
         ULONG status = TraceSetInformation(
             trace_.registrationHandle_, 
-            informationClass, 
-            traceInformation, 
-            informationLength);
+            information_class,
+            trace_information,
+            information_length);
 
         error_check_common_conditions(status);
     }
--- a/krabs/krabs/etw.hpp
+++ b/krabs/krabs/etw.hpp
@@ -79,6 +79,16 @@ namespace krabs { namespace details {
 
         /**
          * <summary>
+         * Configures the ETW trace session settings.
+         * </summary>
+         */
+        void set_trace_information(
+            TRACE_INFO_CLASS informationClass,
+            PVOID traceInformation,
+            ULONG informationLength);
+
+        /**
+         * <summary>
          * Notifies the underlying trace of the buffers that were processed.
          * </summary>
          */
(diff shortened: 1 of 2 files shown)


krabs/krabs/etw.hpp
```diff
@@ -77,6 +77,16 @@ namespace krabs { namespace details {
          */
         EVENT_TRACE_PROPERTIES query();
 
+        /**
+         * <summary>
+         * Configures the ETW trace session settings.
+         * </summary>
```
swannman, line 83: let's update the comment with a link to https://docs.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-tracesetinformation so that users know how to configure these parameters

krabs/krabs/etw.hpp
```diff
@@ -77,6 +77,16 @@ namespace krabs { namespace details {
...
+         */
+        void set_trace_information(
+            TRACE_INFO_CLASS informationClass,
```
swannman, line 86: let's use snake_case here to match the style of the surrounding code: `information_class`, `trace_information`, `information_length`

krabs/krabs/etw.hpp
```diff
@@ -168,7 +178,9 @@ namespace krabs { namespace details {
     template <typename T>
     EVENT_TRACE_LOGFILE trace_manager<T>::open()
     {
-        register_trace();
+        if (trace_.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE) {
```
swannman, line 181: is this change needed?

krabs/krabs/etw.hpp
```diff
@@ -185,6 +197,25 @@ namespace krabs { namespace details {
         return query_trace();
     }
 
+    template <typename T>
+    void trace_manager<T>::set_trace_information(
+        TRACE_INFO_CLASS informationClass,
+        PVOID traceInformation,
+        ULONG informationLength)
+    {
+        if (trace_.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE) {
```
swannman, line 206: is this check and call needed? my preference would be to wrap TraceSetInformation as closely as possible without doing extra work on behalf of the caller, like calling `register_trace` if they haven't yet

krabs/krabs/etw.hpp
```diff
@@ -185,6 +197,25 @@ namespace krabs { namespace details {
         return query_trace();
     }
 
+    template <typename T>
+    void trace_manager<T>::set_trace_information(
+        TRACE_INFO_CLASS informationClass,
```
swannman, line 202: let's use snake_case for these three parameter names per the comment in etw.hpp

krabs/krabs/trace.hpp
```diff
@@ -123,6 +123,30 @@ namespace krabs {
...
+         *    event_id.EventGuid = krabs::guids::perf_info;
+         *    event_id.Type = 46; // SampleProfile
+         *    trace_.set_trace_information(TraceStackTracingInfo, &event_id, sizeof(STACK_TRACING_EVENT_ID));
+         *    krabs::kernel_provider stack_walk_provider(EVENT_TRACE_FLAG_PROFILE, krabs::guids::stack_walk);
+         *    trace_.enable(stack_walk_provider);
+         *    trace.start();
+         * </example>
+         */
+        void set_trace_information(
+            TRACE_INFO_CLASS informationClass,
```
swannman, line 146: let's use snake_case for these parameter names

krabs/krabs/trace.hpp
```diff
@@ -319,6 +343,16 @@ namespace krabs {
         properties_.LogFileMode = properties->LogFileMode;
     }
 
+    template <typename T>
+    void trace<T>::set_trace_information(
+        TRACE_INFO_CLASS informationClass,
```
swannman, line 348: let's use snake_case for these parameter names

swannman (changes requested):
Looking good! Added a few minor observations.

krabs/krabs/etw.hpp
pierricgimmig: 'set_trace_information` needs the session handle, which is currently assigned by the blocking call to 'trace::start()'. This is impractical as we call 'trace::start()' in its own thread. We would then have to somehow synchronize with the assignment of the session handle before safely calling 'set_trace_information'. At the root, the fact 'trace::start' is responsible for both calls to 'StartTrace' and 'OpenTrace' is the issue here. The proposed change allows to split those calls if needed by calling `StartTrace`, which assigns the session handle, before the blocking call to 'trace::start()'. Am I missing something?

krabs/krabs/etw.hpp
pierricgimmig: Done.

krabs/krabs/trace.hpp
pierricgimmig: Done.

krabs/krabs/trace.hpp
pierricgimmig: Done.

krabs/krabs/etw.hpp
pierricgimmig: Done.

krabs/krabs/etw.hpp
pierricgimmig: Done.

krabs/krabs/etw.hpp
swannman: Got it! Rather than adding a call to `register_trace` here, would this sequence of calls give you the outcome you're looking for?

```
krabs::trace trace;
krabs::guid id(L"{A0C1853B-5C40-4B15-8766-3CF1C58F985A}");
provider<> powershell(id);
trace.enable(powershell);
trace.open(); // register_trace is called in here
trace.set_trace_information(...); // new
trace.process(); // this blocks
```

pierricgimmig: Ah yes, I was indeed missing something :-). Thanks, yep this is perfect, I verified that it works. I removed the checks and modified the function comment for `set_trace_information`.

swannman (approved):
Looks great!

swannman:
This change will be included in the 4.2.1 nuget version which I am publishing today.

pierricgimmig:
> This change will be included in the 4.2.1 nuget version which I am publishing today.

Awesome, thanks a lot for your help, @swannman !

Issue #26, opened 2017-08-30
how to get the stack info when receive evnet?


chena1982:
It seems need set EVENT_ENABLE_PROPERTY_STACK_TRACE when call EnabeTraceEx2.
but i can't see any api to enbale this.
Thanks.

zacbrown:
Hi,

Could you provide some sample code and an explanation for what you're trying to do?

Thanks,

Zac

chena1982:
```
	PEVENT_FILTER_EVENT_ID filterId;
	DWORD size = FIELD_OFFSET(EVENT_FILTER_EVENT_ID, Events[2]);
	filterId = (PEVENT_FILTER_EVENT_ID)new BYTE[size];
	filterId->FilterIn = TRUE;
	filterId->Count = 2;
	filterId->Events[0] = 0x2b;
	filterId->Events[1] = 0x21;


	EVENT_FILTER_DESCRIPTOR filterDesc;
	ZeroMemory(&filterDesc, sizeof(filterDesc));
	filterDesc.Ptr = (ULONGLONG)filterId;
	filterDesc.Size = size;
	filterDesc.Type = EVENT_FILTER_TYPE_STACKWALK;

	ENABLE_TRACE_PARAMETERS traceParam;
	ZeroMemory(&traceParam, sizeof(traceParam));
	traceParam.Version = ENABLE_TRACE_PARAMETERS_VERSION_2;
	traceParam.EnableProperty = EVENT_ENABLE_PROPERTY_STACK_TRACE;
	traceParam.SourceId = guid[(int)type];
	traceParam.EnableFilterDesc = &filterDesc;
	traceParam.FilterDescCount = 1;


	_status = EnableTraceEx2(hSession, &guid[(int)type], EVENT_CONTROL_CODE_ENABLE_PROVIDER, level, anyKeyword, allKeyword, 0, &traceParam);
```

some code to enable stack trace info.

zacbrown:
Hi,

We can look into making the parameters more configurable. I'll make a bug. It will likely be a month before we make the change as we finish out our planning cycle for this quarter.

-Zac

zacbrown:
I've implemented the ability to set  EVENT_ENABLE_PROPERTY_STACK_TRACE for the trace_flags on a given provider object. I did not expose this functionality in the .NET layer. It's only available in the native krabsetw API.

[...]

PR #191, opened 2022-04-11
add stack trace helpers (#26)

jdu2600:
I've added some helper functions for parsing stack traces from a record's extended data.

In particular, this exposes stack traces to C#.

### DIFF
--- a/Microsoft.O365.Security.Native.ETW/IEventRecord.hpp
+++ b/Microsoft.O365.Security.Native.ETW/IEventRecord.hpp
@@ -402,6 +402,12 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         /// <param name="result">the resulting byte array</param>
         /// <returns>true if fetching the data succeeded, false otherwise</returns>
         bool TryGetBinary(String^ name, [Out] array<Byte>^% result);
+
+        /// <summary>
+        ///  Retrieves the call stack associated with the record, if enabled.
+        /// </summary>
+        /// <returns>a list of return addresses</returns>
+        List<UIntPtr>^ StackTrace();
     };
 
 } } } }
--- a/Microsoft.O365.Security.Native.ETW/IEventRecord.hpp
+++ b/Microsoft.O365.Security.Native.ETW/IEventRecord.hpp
@@ -407,7 +407,7 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         ///  Retrieves the call stack associated with the record, if enabled.
         /// </summary>
         /// <returns>a list of return addresses</returns>
-        List<UIntPtr>^ StackTrace();
+        List<UIntPtr>^ GetStackTrace();
     };
 
 } } } }
--- a/krabsetw.nuspec
+++ b/krabsetw.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Krabsetw</id>
-        <version>4.2.2</version>
+        <version>4.2.3</version>
         <title>Krabs ETW Wrappers</title>
         <authors>Microsoft</authors>
         <owners>Microsoft</owners>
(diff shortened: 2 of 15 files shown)


jdu2600:
cc: @pathtofile

jdu2600:
🤔 The CI Build failure appears unrelated to this PR?

pathtofile:
Awesome Idea!
As for the CI break, it's because it's using `windows-latest`, which switched to only having Visual Studio 2022 on it.
I fixed this with Sealighter's CI using the `windows-2019` target: https://github.com/pathtofile/Sealighter/blob/main/.github/workflows/main.yml

jdu2600:
CI fixed. That was easy. 😁 
I need to learn github workflows  one day.

Hmm. So how do I credit you on this PR?
I pretty much just wrapped the code you pasted in the ticket. 😃

krabsetw.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Krabsetw</id>
-        <version>4.2.2</version>
```
swannman, line 5: Thanks for updating the nuspec! Would you be willing to update `O365.Security.Native.ETW.Debug.nuspec` and `O365.Security.Native.ETW.nuspec` as well?

Microsoft.O365.Security.Native.ETW/IEventRecord.hpp
```diff
@@ -402,6 +402,12 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
         /// <param name="result">the resulting byte array</param>
         /// <returns>true if fetching the data succeeded, false otherwise</returns>
         bool TryGetBinary(String^ name, [Out] array<Byte>^% result);
+
+        /// <summary>
+        ///  Retrieves the call stack associated with the record, if enabled.
+        /// </summary>
+        /// <returns>a list of return addresses</returns>
+        List<UIntPtr>^ StackTrace();
```
swannman, line 410: optional: I think I'd prefer `GetStackTrace()` here since this is a method rather than a property.

swannman (changes requested):
Thanks for doing this, looks great! Added a couple of minor comments.

swannman:
@jdu2600 just a quick ping - minor code review feedback and then this is good to check in!

krabsetw.nuspec
jdu2600: I think I've already done this?

https://github.com/microsoft/krabsetw/blob/e1d416ecac2a58fcaec531fda801a45d69c2c7f9/O365.Security.Native.ETW.Debug.nuspec#L4-L5
https://github.com/microsoft/krabsetw/blob/e1d416ecac2a58fcaec531fda801a45d69c2c7f9/O365.Security.Native.ETW.nuspec#L4-L5

jdu2600:
Just back from vacation. 😄

Microsoft.O365.Security.Native.ETW/IEventRecord.hpp
jdu2600: Would you prefer the C++ version be renamed to `get_stack_trace()` also?

Also, I currently wrap the stack trace in a `std::vector` - but I could instead return a pointer to the data in the record and let the caller copy if required? Or perhaps even expose both options?

krabsetw.nuspec
swannman: You're right, my apologies! I saw krabsetw.nuspec at the bottom of the changelist and didn't scroll up to find the others.

swannman (approved):
Looks great!

Microsoft.O365.Security.Native.ETW/IEventRecord.hpp
swannman: The C++ naming seems idiomatic as-is, so let's leave it. Since `EVENT_EXTENDED_ITEM_STACK_TRACE64` seems to only be a list of addresses I think a `std::vector` of them is a good approach.

PR #206, opened 2023-03-29
Add trace parsing from a file instead of realtime

acyr:
Add a new interface set_trace_filename() that will enable reading from an ETL file rather than real-time capture.
Also added a new NativeExample that demonstrates this capability with a sample ETL.


acyr:

examples/NativeExamples/examples.h
```diff
@@ -8,6 +8,11 @@ struct kernel_and_user_trace_001
     static void start();
 };
 
+struct user_trace_009_from_file
```
swannman, line 11: let's move this to the bottom of the file

examples/NativeExamples/NativeExamples.vcxproj
```diff
@@ -152,6 +152,7 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
+    <ClCompile Include="user_trace_009_from_file.cpp" />
```
swannman, line 155: let's move this to the line after `user_trace_008_stacktrace.cpp` below

examples/NativeExamples/user_trace_009_from_file.cpp
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// This example shows how to use a user_trace and a kernel_trace in the same program.
```
swannman, line 4: let's update this comment to reflect what this file is demonstrating

examples/NativeExamples/user_trace_009_from_file.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <iostream>
+#include <thread>
+#include <condition_variable>
+#include "..\..\krabs\krabs.hpp"
+#include "examples.h"
+
+static void setup_ps_provider(krabs::provider<>& provider);
```
swannman, line 12: are either of these being used?

examples/NativeExamples/user_trace_009_from_file.cpp
```diff
@@ -0,0 +1,67 @@
...
+        }
+        });
+
+    // the user_trace needs to know about the provider that we've set up.
+    trace.enable(provider);
+
+    // specify a filename to read from, will disable realtime and source from file instead
+    trace.set_trace_filename(L"..\\..\\examples\\NativeExamples\\powershell.etl");
+
+    // begin listening for events. This call blocks, so if you want to do other things
```
swannman, line 64: what happens when the end of file is reached - does the trace session continue to block the thread?

krabs/krabs/etw.hpp
```diff
@@ -262,9 +262,18 @@ namespace krabs { namespace details {
     EVENT_TRACE_LOGFILE trace_manager<T>::fill_logfile()
     {
         EVENT_TRACE_LOGFILE file = {};
-        file.LoggerName          = const_cast<wchar_t*>(trace_.name_.c_str());
-        file.ProcessTraceMode    = PROCESS_TRACE_MODE_EVENT_RECORD |
-                                   PROCESS_TRACE_MODE_REAL_TIME;
+
+        if (!trace_.logFilename_.empty())
```
swannman, line 266: nit: let's maintain the original alignment on the `=` characters:

```
        if (!trace_.logFilename_.empty())
        {
            file.LogFileName         = const_cast<wchar_t*>(trace_.logFilename_.c_str());
            file.ProcessTraceMode    = PROCESS_TRACE_MODE_EVENT_RECORD;
        }
        else
        {
            file.LoggerName          = const_cast<wchar_t*>(trace_.name_.c_str());
            file.ProcessTraceMode    = PROCESS_TRACE_MODE_EVENT_RECORD |
                                       PROCESS_TRACE_MODE_REAL_TIME;
        }
```

swannman (changes requested):
Thanks for submitting this! Added a few comments, mostly style-related.

examples/NativeExamples/user_trace_009_from_file.cpp
acyr: It unblocks and returns at EOF.  I can call stop, but I don't know if that's necessary... maybe better form?

krabs/krabs/etw.hpp
acyr: Will do.

examples/NativeExamples/user_trace_009_from_file.cpp
acyr: Whoops.  Nope, leftover garbage.  Originally user kernel_and_user_trace_001 as the base for the file.

examples/NativeExamples/user_trace_009_from_file.cpp
acyr: I'll add a stop() call for completeness, and update comment to reflect the EOF results in start() returning control.

swannman (approved):

swannman:
@acyr do you want to bump the version number in the three `.nuspec` files in the root of the project so we can publish updated nugets?

swannman (approved):

Issue #214, opened 2023-09-13
Add out type for the property (_TDH_OUT_TYPE)


kaaleksandr:
class **property** contains the **type** method, it is necessary to add the **out_type** method to output the output type of the property

swannman:
Resolved by #215. Thank you @kaaleksandr!

PR #215, opened 2023-09-14
#214

kaaleksandr:
Added a property for out type output

### DIFF
--- a/krabs/krabs/property.hpp
+++ b/krabs/krabs/property.hpp
@@ -48,7 +48,7 @@ namespace krabs {
          *   object do this for you with its `properties` method.
          * </remarks>
          */
-        property(const std::wstring &name, _TDH_IN_TYPE type);
+        property(const std::wstring &name, _TDH_IN_TYPE type, _TDH_OUT_TYPE outType);
 
         /**
          * <summary>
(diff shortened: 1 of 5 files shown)


krabs/krabs/property.hpp
```diff
@@ -64,9 +64,17 @@ namespace krabs {
          */
         _TDH_IN_TYPE type() const;
 
+        /**
+         * <summary>
+         * Retrieves the Tdh type of the property.
```
swannman, line 69: consider `Retrieves the Tdh out type of the property.` to disambiguate `out_type` from `type` above

swannman (changes requested):
Looks good, I added one minor comment. If you need updated nugets published so you can consume this change, would you mind bumping the nuget versions in the three `.nuspec` files at the root of the repo? Thanks!

kaaleksandr:
> Looks good, I added one minor comment. If you need updated nugets published so you can consume this change, would you mind bumping the nuget versions in the three `.nuspec` files at the root of the repo? Thanks!

Yes of course

swannman (approved):

swannman:
@kaaleksandr ready to merge once you accept the CLA above. Thanks!

kaaleksandr:

PR #226, opened 2024-01-19
Stability fixes in native code

starix:
Some exceptions hanling + buffer overfow fix

### DIFF
--- a/krabs/krabs/etw.hpp
+++ b/krabs/krabs/etw.hpp
@@ -202,6 +202,9 @@ namespace krabs { namespace details {
         PVOID trace_information,
         ULONG information_length)
     {
+        if (trace_.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE)
+            return;
+
         ULONG status = TraceSetInformation(
             trace_.registrationHandle_, 
             information_class,
@@ -283,12 +286,16 @@ namespace krabs { namespace details {
     template <typename T>
     void trace_manager<T>::stop_trace()
     {
+        if (trace_.registrationHandle_ == INVALID_PROCESSTRACE_HANDLE)
+            return;
+
         trace_info info = fill_trace_info();
         ULONG status = ControlTrace(
             NULL,
             trace_.name_.c_str(),
             &info.properties,
             EVENT_TRACE_CONTROL_STOP);
+        trace_.registrationHandle_ = INVALID_PROCESSTRACE_HANDLE;
 
         if (status != ERROR_WMI_INSTANCE_NOT_FOUND) {
             error_check_common_conditions(status);
@@ -328,6 +335,7 @@ namespace krabs { namespace details {
                 status = StartTrace(&trace_.registrationHandle_,
                     trace_.name_.c_str(),
                     &info.properties);
+                error_check_common_conditions(status);
             }
             catch (need_to_be_admin_failure) {
                 (void)open_trace();
@@ -344,6 +352,15 @@ namespace krabs { namespace details {
                 // In some versions, the error code is 87 when using
                 // SystemTraceControlGuid session. If open/close doesn't
                 // throw, then we can continually processing events.
+                (void)open_trace();
+                close_trace();
+                status = ERROR_SUCCESS;
+                trace_.registrationHandle_ = INVALID_PROCESSTRACE_HANDLE;
+            }
+            catch (trace_already_registered) {
+                // We can't StartTrace for "Microsoft-Windows-Security-Auditing"
+                // provider. If open/close doesn't throw, then we can continually
+                // processing events.
                 (void)open_trace();
                 close_trace();
                 status = ERROR_SUCCESS;
--- a/krabs/krabs/testing/extended_data_builder.hpp
+++ b/krabs/krabs/testing/extended_data_builder.hpp
@@ -96,7 +96,7 @@ namespace krabs { namespace testing {
         // No null terminator
         BYTE guid_data[GUID_STRING_LENGTH_NO_BRACES] = {};
 
-        StringFromGUID2(container_id, wide_guid_buffer, GUID_STRING_LENGTH_WITH_BRACES + 1);
+        StringFromGUID2(container_id, wide_guid_buffer, sizeof(wide_guid_buffer));
 
         for (int i = 0; i < GUID_STRING_LENGTH_NO_BRACES; i++)
         {
--- a/krabs/krabs/trace.hpp
+++ b/krabs/krabs/trace.hpp
@@ -282,6 +282,12 @@ namespace krabs {
          */
         void set_default_event_callback(c_provider_callback callback);
 
+        template <typename U>
+        void set_default_event_callback(U& callback);
+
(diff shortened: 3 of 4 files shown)


krabs/krabs/testing/extended_data_builder.hpp
```diff
@@ -96,7 +96,7 @@ namespace krabs { namespace testing {
         // No null terminator
         BYTE guid_data[GUID_STRING_LENGTH_NO_BRACES] = {};
 
-        StringFromGUID2(container_id, wide_guid_buffer, GUID_STRING_LENGTH_WITH_BRACES + 1);
+        StringFromGUID2(container_id, wide_guid_buffer, sizeof(wide_guid_buffer));
```
kylereedmsft, line 99: This isn't correct. The third parameter should be the count of characters not count of bytes. What symptom were you seeing that made change this?

krabs/krabs/trace.hpp
```diff
@@ -348,7 +354,8 @@ namespace krabs {
     template <typename T>
     trace<T>::~trace()
     {
-        stop();
+        // Never throw in destructor
+        try { stop(); } catch (...) {}
```
kylereedmsft, line 358: Did you see this throw? I'd need to look to see how trace_manager can throw.
Usually, we don't want to catch everything. There's also implications for SEH callbacks that might have been set up.

krabs/krabs/ut.hpp
```diff
@@ -164,7 +164,7 @@ namespace krabs { namespace details {
             std::vector<BYTE> filterEventIdBuffer;
             auto filterEventIdCount = settings.provider_filter_event_ids_.size();
 
-            if (filterEventIdCount > 0) {
+            if (filterEventIdCount > 0 && filterEventIdCount <= MAX_EVENT_FILTER_EVENT_ID_COUNT) {
```
kylereedmsft, line 167: This should be an error condition that is bubbled up to the caller and then we can either:
* Return an error condition (throw or return an error code)
* Use the first MAX_EVENT_FILTER_EVENT_ID_COUNT items

I prefer throwing so it's obvious what went wrong.

krabs/krabs/ut.hpp
```diff
@@ -237,15 +237,18 @@ namespace krabs { namespace details {
...
+        {
+            auto eventInfo = trace.context_.schema_locator.get_event_schema(record);
+            for (auto& provider : trace.providers_) {
+                if (eventInfo->ProviderGuid == provider.get().guid_) {
+                    provider.get().on_event(record, trace.context_);
+                    return;
+                }
             }
         }
+        catch (krabs::could_not_find_schema&) {}
```
kylereedmsft, line 251: This exception should not be caught. How else will a caller know why the callback is not working?
This layer needs to bubble up errors so that the calling application can know something went wrong.

krabs/krabs/trace.hpp
```diff
@@ -440,7 +447,30 @@ namespace krabs {
...
     void trace<T>::set_default_event_callback(c_provider_callback callback)
     {
+        // C function pointers don't interact well with std::ref, so we
+        // overload to take care of this scenario.
         default_callback_ = callback;
     }
 
+    template <typename T>
+    template <typename U>
+    void trace<T>::set_default_event_callback(U& callback)
```
kylereedmsft, line 457: Do you have an example of what you're trying to solve here? I don't think you should need these at all.
If there is a good reason to have this, you should use a single function:

```cpp
    template <typename T>
    template <typename U>
    void trace<T>::set_default_event_callback(U&& callback)
```

This can bind to all flavors of parameters correctly.

kylereedmsft (changes requested):
Thanks for your PR. I have a few questions about your changes.

krabs/krabs/testing/extended_data_builder.hpp
starix: My mistake, rolling back

krabs/krabs/etw.hpp
```diff
@@ -349,6 +357,15 @@ namespace krabs { namespace details {
                 status = ERROR_SUCCESS;
                 trace_.registrationHandle_ = INVALID_PROCESSTRACE_HANDLE;
             }
+            catch (trace_already_registered) {
+                // We can't StartTrace for "Microsoft-Windows-Security-Auditing"
+                // provider. If open/close doesn't throw, then we can continually
```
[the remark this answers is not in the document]
starix: In some rare cases StartTrace() return ERROR_ALREADY_EXISTS. ControlTrace(EVENT_TRACE_CONTROL_STOP) inside stop_trace() return ERROR_SUCCESS, but second call of StartTrace() give us ERROR_ALREADY_EXISTS again.
It's rare case so let's rollback now. But I will try to return with POC.

krabs/krabs/trace.hpp
starix: Throwing an exception out of a destructor is dangerous. If another exception is already propagating the application will terminate.
ControlTrace(EVENT_TRACE_CONTROL_STOP) or CloseTrace() can return error inside stop(). Using SEH is to much.

krabs/krabs/ut.hpp
starix: Both variants not ideal. We limit the user's use of krabs filtering. In my case we just "turn off" system prefiltration. We will get all events from provider but have a chance to filter them from in callback.

krabs/krabs/ut.hpp
starix: And trace session will close without any chance to resume. In my case we can still handle event in default callback.

krabs/krabs/trace.hpp
kylereedmsft: I'm aware that destructors shouldn't throw, however it's not a good pattern to just add "catch all" blocks in all the destructors. Catch what we expect.

krabs/krabs/trace.hpp
starix: > Do you have an example of what you're trying to solve here?
```
auto fun = [&](const EVENT_RECORD& record) {  };
krabs::user_trace trace;
trace.set_default_event_callback(fun);

```
The same approach is used in `event_filter::add_on_error_callback` or `base_provider<T>::add_on_event_callback`

krabs/krabs/trace.hpp
starix: Ok. If we add error_callback to trace object it can solve a problem with user notification.

granmaa: I have observed `trace<T>::~trace()` throwing:
`
_CxxThrowException
krabs::error_check_common_conditions+0x9c
krabs::details::trace_manager<krabs::trace<krabs::details::ut> >::stop_trace
krabs::details::trace_manager<krabs::trace<krabs::details::ut> >::stop
krabs::trace<krabs::details::ut>::stop
krabs::trace<krabs::details::ut>::~trace<krabs::details::ut>
`
with status == 4 (`ERROR_TOO_MANY_OPEN_FILES`)

PR #231, opened 2024-03-18
Fix errors in clang

gmh5225:


gmh5225:

swannman (approved):

PR #250, opened 2024-11-12
Only allocate one EventRecordMetadata^ and reuse it for RawProvider events

mjeong92:
RawProvider::EventNotification created a new managed object EventRecordMetadata for each EVENT_RECORD it received. Since each provider is single threaded, we should be able to reuse a single allocation by replacing the pointers to the user data in EVENT_RECORD native object on a single EventRecordMetadata before raising the event up to the managed layer.

This change makes the user record and the header pointers in EventRecordMetadata accessible internally so that they can be updated from each EVENT_RECORD received in the RawProvider. RawProvider adds a EventRecordMetadata^ member that gets initialized in the constructor and re-used for each EVENT_RECORD by updating the two pointers each time.

### DIFF
--- a/Microsoft.O365.Security.Native.ETW/EventRecordMetadata.hpp
+++ b/Microsoft.O365.Security.Native.ETW/EventRecordMetadata.hpp
@@ -17,14 +17,15 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
     /// </summary>
     public ref class EventRecordMetadata : public IEventRecordMetadata
     {
-    protected:
-        const EVENT_RECORD* record_;
-        const EVENT_HEADER* header_;
-
     internal:
         EventRecordMetadata(const EVENT_RECORD& record)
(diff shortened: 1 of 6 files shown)


Microsoft.O365.Security.Native.ETW/EventRecordMetadata.hpp
```diff
@@ -26,6 +26,17 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
             : record_(&record)
             , header_(&record.EventHeader) { }
 
+        EventRecordMetadata() { }
+
+        /// <summary>
+        /// Update the underlying pointers to the event data to the given record.
```
kylereedmsft, line 32: `Updates this instance to point to the specified event record.`

kylereedmsft (dismissed):

kylereedmsft (approved):

Issue #253, opened 2025-01-28
Is schema valid for WPP event?


wwh1004:
The schema told me that the WPP_ThisDir_CTLGUID_OLE32 WPP event has these properties. But in fact the value of the property is completely wrong.

Is my understanding of the schema correct, or is the schema invalid for WPP events?

![Image](https://github.com/user-attachments/assets/560e34ec-218f-4792-abba-d9d34821dfca)

Issue #252, opened 2025-01-28
could_not_find_schema exception was thrown when stop tracing


wwh1004:
In kernel_trace_002.cpp, I added the stop code. When stopping, a could_not_find_schema exception will be thrown. I tested the newly installed, latest Windows 11 preview version, and the same exception occurred.

```
Exception caught: Could not find the schema: status_code=1168 provider_id=01853a65-418f-4f36-aefc-dc0f1d2fd235 event_id=0
```

provider: 01853a65-418f-4f36-aefc-dc0f1d2fd235 opcode: 0x13

![Image](https://github.com/user-attachments/assets/fecd0b4d-b483-47d0-90a5-d211436f2818)

I believe this opcode is not registered with MOF, at least I can't find it in EtwExplorer.

![Image](https://github.com/user-attachments/assets/965c16d1-a655-481b-959e-e607a674efcf)

``` cpp
#include <iostream>
#include "..\..\krabs\krabs.hpp"
#include "..\..\krabs\krabs\perfinfo_groupmask.hpp"
#include "examples.h"
#include <thread>

void kernel_trace_002::start()
{
    krabs::kernel_trace trace(L"kernel_trace_002");

    krabs::kernel::object_manager_provider ob_provider;
    ob_provider.add_on_event_callback([](const EVENT_RECORD& record, const krabs::trace_context& trace_context) {
        if (record.EventHeader.EventDescriptor.Opcode == 33) {
            krabs::schema schema(record, trace_context.schema_locator);
            krabs::parser parser(schema);
            std::wstring name = parser.parse<std::wstring>(L"ObjectName");
            if (name.length() >= 3 && name.compare(name.length() - 3, 3, L"dll") == 0)
                std::wcout << L"Handle closed for object with name " << name << std::endl;
        }
        });
    trace.enable(ob_provider);

    krabs::kernel_provider hive_provider(GUID_NULL, PERF_REG_HIVE);

    trace.set_default_event_callback([](const EVENT_RECORD& record, const krabs::trace_context& trace_context) {
        krabs::schema schema(record, trace_context.schema_locator);
        std::wcout << std::to_wstring(record.EventHeader.ProviderId);
        std::wcout << L" provider=" << schema.provider_name();
        std::wcout << L" event_name=" << schema.event_name();
        std::wcout << L" task_name=" << schema.task_name();
        std::wcout << L" opcode=" << schema.event_opcode();
        std::wcout << L" opcode_name=" << schema.opcode_name();
[...]

PR #255, opened 2025-02-01
Unify. NET wrappers of callbacks and more optimizations

wwh1004:
1. Unify. NET wrappers of callbacks to simplify logic
2. Add no throw version of get_event_schema and also cache error status
3. Merge the ability of RawProvider to Provider to reduce the fragmentation of code functionality
4. Remove invalid GCHandle usages
5. Use schema_locator.get_event_schema_no_throw in ut::forward_events to avoid exceptions
6. Check status code instead of throw then catch exceptions to gain more performance
7. Also cache EventRecord and EventRecordError objects like EventRecordMetadata to reduce allocations

Closes https://github.com/microsoft/krabsetw/issues/252
Closes https://github.com/microsoft/krabsetw/issues/253

### DIFF
--- a/Microsoft.O365.Security.Native.ETW/Callbacks.hpp
+++ b/Microsoft.O365.Security.Native.ETW/Callbacks.hpp
@@ -1,0 +1,154 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+
+#include <krabs.hpp>
+
+#include "EventRecord.hpp"
+#include "EventRecordError.hpp"
+
+using namespace System;
+using namespace System::Runtime::InteropServices;
+
+namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
+
+    /// <summary>
+    /// Delegate called when a new ETW <see cref="O365::Security::ETW::EventRecord"/> is received.
+    /// </summary>
+    public delegate void IEventRecordDelegate(IEventRecord^ record);
+
+    /// <summary>
+    /// Delegate called when a new ETW <see cref="O365::Security::ETW::EventRecordMetadata"/> is received.
+    /// </summary>
+    public delegate void IEventRecordMetadataDelegate(IEventRecordMetadata^ record);
+
+    /// <summary>
+    /// Delegate called on errors when processing an <see cref="O365::Security::ETW::EventRecord"/>.
+    /// </summary>
+    public delegate void EventRecordErrorDelegate(IEventRecordError^ error);
+
+    delegate void EventNativeDelegate(const EVENT_RECORD&, const krabs::trace_context&);
+
+    delegate void EventErrorNativeDelegate(const EVENT_RECORD&, const std::string&);
+
+    ref class CallbackBridge {
+    private:
+
+        EventNativeDelegate^ eventDelegateKeepAlive_;
+        EventErrorNativeDelegate^ errorDelegateKeepAlive_;
+        EventRecordMetadata^ metadata_;
+        EventRecord^ record_;
+        EventRecordError^ error_;
+
+        EventRecordMetadata^ CreateMetadata(const EVENT_RECORD& record);
+        EventRecord^ CreateRecord(const EVENT_RECORD& record, const krabs::schema& schema, krabs::parser& parser);
+        EventRecordError^ CreateError(const std::string& error_message, const EVENT_RECORD& record);
+
+        void EventNotification(const EVENT_RECORD& record, const krabs::trace_context& trace_context);
+        void ErrorNotification(const EVENT_RECORD& record, const std::string& error_message);
+
+    public:
+        IEventRecordMetadataDelegate^ OnMetadata;
+        IEventRecordDelegate^ OnEvent;
+        EventRecordErrorDelegate^ OnError;
+
+        krabs::c_provider_callback GetOnEventBridge()
+        {
+            if (!eventDelegateKeepAlive_)
+                eventDelegateKeepAlive_ = gcnew EventNativeDelegate(this, &CallbackBridge::EventNotification);
+            return (krabs::c_provider_callback)Marshal::GetFunctionPointerForDelegate(eventDelegateKeepAlive_).ToPointer();
+        }
+
+        krabs::c_provider_error_callback GetOnErrorBridge()
+        {
+            if (!errorDelegateKeepAlive_)
+                errorDelegateKeepAlive_ = gcnew EventErrorNativeDelegate(this, &CallbackBridge::ErrorNotification);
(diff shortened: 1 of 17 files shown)


krabs/krabs/ut.hpp
```diff
@@ -238,12 +238,14 @@ namespace krabs { namespace details {
         // for MOF providers, EventHeader.Provider is the *Message* GUID
         // we need to ask TDH for event information in order to determine the
         // correct provider to pass this event to
-        auto schema = get_event_schema_from_tdh(record);
-        auto eventInfo = reinterpret_cast<PTRACE_EVENT_INFO>(schema.get());
-        for (auto& provider : trace.providers_) {
-            if (eventInfo->ProviderGuid == provider.get().guid_) {
-                provider.get().on_event(record, trace.context_);
-                return;
+        TDHSTATUS status = ERROR_SUCCESS;
```
kylereedmsft, line 241: Before this may have thrown but now it will not. That seems like a big difference in behavior. Need to understand if anything downstream expected this to be throwing. That said, it's _probably_ better this way, but I wonder if there are any cases for TDHSTATUS where it would make the most sense to fail here.

Microsoft.O365.Security.Native.ETW/Callbacks.hpp
```diff
@@ -0,0 +1,154 @@
...
+    delegate void EventNativeDelegate(const EVENT_RECORD&, const krabs::trace_context&);
+
+    delegate void EventErrorNativeDelegate(const EVENT_RECORD&, const std::string&);
+
+    ref class CallbackBridge {
+    private:
+
+        EventNativeDelegate^ eventDelegateKeepAlive_;
+        EventErrorNativeDelegate^ errorDelegateKeepAlive_;
+        EventRecordMetadata^ metadata_;
```
kylereedmsft, line 40: NB: these instances are reused which reduces per-event allocations.

Microsoft.O365.Security.Native.ETW/Callbacks.hpp
```diff
@@ -0,0 +1,154 @@
...
+    public:
+        IEventRecordMetadataDelegate^ OnMetadata;
+        IEventRecordDelegate^ OnEvent;
+        EventRecordErrorDelegate^ OnError;
+
+        krabs::c_provider_callback GetOnEventBridge()
+        {
+            if (!eventDelegateKeepAlive_)
+                eventDelegateKeepAlive_ = gcnew EventNativeDelegate(this, &CallbackBridge::EventNotification);
+            return (krabs::c_provider_callback)Marshal::GetFunctionPointerForDelegate(eventDelegateKeepAlive_).ToPointer();
```
kylereedmsft, line 60: Nice, GetFunctionPointerForDelegate builds a stub that doesn't get moved so no pinning is needed.

Microsoft.O365.Security.Native.ETW/Callbacks.hpp
```diff
@@ -0,0 +1,154 @@
...
+
+        krabs::c_provider_error_callback GetOnErrorBridge()
+        {
+            if (!errorDelegateKeepAlive_)
+                errorDelegateKeepAlive_ = gcnew EventErrorNativeDelegate(this, &CallbackBridge::ErrorNotification);
+            return (krabs::c_provider_error_callback)Marshal::GetFunctionPointerForDelegate(errorDelegateKeepAlive_).ToPointer();
+        }
+    };
+
+    EventRecordMetadata^ CallbackBridge::CreateMetadata(const EVENT_RECORD& record)
```
kylereedmsft, line 71: nit: I don't like "Create" for the verb here because the instances are reused. I think I'd be less concerned if they weren't public. Maybe "Wrap" or maybe even "Update" would be better?

Microsoft.O365.Security.Native.ETW/Callbacks.hpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        auto onError = OnError;
+        if (onError) {
+            auto error = CreateError(error_message, record);
+
+            onError(error);
+        }
+    }
+
+#define CombineOrRemoveDelegate(type, target, value, operation) \
```
kylereedmsft, line 140: I'd prefer if this were a template instead of a macro.

kylereedmsft (changes requested):
This PR looks really good. Thank you for the work.
I have a few comments, and I'd like you to change the macro to be a template function instead.

Microsoft.O365.Security.Native.ETW/ITrace.hpp
```diff
@@ -52,11 +52,15 @@ namespace Microsoft { namespace O365 { namespace Security { namespace ETW {
...
         void Enable(Provider^ provider);
 
+#pragma warning(push)
+#pragma warning(disable: 4947) // Deprecated warning
         /// <summary>
         /// Enables a raw provider for the given user trace.
         /// </summary>
         /// <param name="provider">The <see cref="O365::Security::ETW::RawProvider"/> to enable.</param>
+        [Obsolete("The RawProvider is deprecated. Use the Provider.OnMetadata event instead.")]
         void Enable(RawProvider^ provider);
```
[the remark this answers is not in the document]
wwh1004: Yes. This PR provides ABI backward compatibility. However, the newly compiled code will produce a warning by the IDE that the API has been deprecated.

krabs/krabs/ut.hpp
wwh1004: It seems that no one can properly catch the exceptions thrown in forward_events, which lead to an interruption in ProcessTrace. If someone needs to catch exceptions out of ProcessTrace, then there indeed will be an impact on compatibility. However, overall I feel that the behavior after PR is more in line with subjective expectations. If necessary, is it required to add a bool flag to ensure compatibility?

Microsoft.O365.Security.Native.ETW/Callbacks.hpp
wwh1004: It seems that these objects have been reused.

``` cpp
if (!eventDelegateKeepAlive_)
    eventDelegateKeepAlive_ = gcnew EventNativeDelegate(this, &CallbackBridge::EventNotification);
```

Microsoft.O365.Security.Native.ETW/Callbacks.hpp
wwh1004: Fixed

Microsoft.O365.Security.Native.ETW/Callbacks.hpp
wwh1004: Fixed

kylereedmsft (dismissed):
Thanks for the PR. These are good improvements.

kylereedmsft (changes requested):
There are merge conflicts with the other PR. Can you pull main, resolve the conflicts and I'll approve. Thanks!

kylereedmsft (approved):

Issue #240, opened 2024-07-14
GetTraceDataProviders in provider.hpp has high CPU usage and takes 2-3 second to return?


rjadidi920:
I noticed that during my function that creates `krabs::provider`s and registers them, I get a high CPU usage which seemed odd.

I traced the problem to the following line of code in `provider.hpp`:

```
    template <typename T>
    provider<T>::provider(const std::wstring &providerName)
            ....
            hr = allProviders->GetTraceDataProviders(NULL);  // This line is causing high cpu usage and takes 2-3 seconds to return. tested in Windows 10 and Windows 7.
```
 
My question is, is there anyway to improve this? Because right now, for every provider that I want to create and register, it takes 2-3 second and high cpu usage because of this, so imagine when I try to register 10 providers..

I wanted to use the GUID version of the provider constructor, but the problem is I don't see it validating the GUID, because i might use providers that exist in some systems but don't exist in others, so I need to check if the GUID provider exist or not to begin with.

PR #256, opened 2025-02-01
Use TdhEnumerateProviders to obtain GUIDs to speed up the conversion and avoid conflicts with CoInitializeEx.

wwh1004:
On my computer, it takes about 5 seconds to use clsid_tracedataproviderCollection, and it only takes 100 milliseconds to use TdhenumerateProvider.
Closes https://github.com/microsoft/krabsetw/issues/240

### DIFF
--- a/krabs/krabs/provider.hpp
+++ b/krabs/krabs/provider.hpp
@@ -16,8 +16,6 @@
 
 #include <evntcons.h>
 #include <guiddef.h>
-#include <pla.h>
-#include <comutil.h>
 
 #ifdef _DEBUG
 #pragma comment(lib, "comsuppwd.lib")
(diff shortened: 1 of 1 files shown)


kylereedmsft:
Seems like a good change. Do we understand if there are any differences between the two?

kylereedmsft (approved):

wwh1004:
The usage of TdhEnumerateProvider comes from https://github.com/microsoft/perfview/tree/main/src/TraceEvent. To be honest, I am not familiar with the details of ITraceDataProviderCollection, but by using IDA analysis, the data source should also be Tdh: `PlaiAllocateBufferAndRead<CPlaiTdhEnumProviders>`.

PR #272, opened 2026-03-09
Update NuGet package version to 4.4.7

vmurthysuhas:
The previous PR missed updating the version to publish a new nuget.


O365.Security.Native.ETW.nuspec
kylereedmsft: Update this too

kylereedmsft (approved):

PR #274, opened 2026-04-01
Parser performance feedback

jdu2600:
Tidy #273

PR #275, opened 2026-04-01
replace name->property map with hinted linear scan

jdu2600:
Builds on #273.
Alternative to #274.

Assuming that callers consume **all** properties (sanely) in event order then the (fallback) hinted scan is the fastest approach.
If the caller consumes properties (perversely) in the reverse order, performance degrades though.
I also didn't benchmark sparse consumption.

| Event Type | Props | last+1<br/>(best/expected) | last+1<br/>(worst case) | map<br/>(4.4.8) | bugged<br/>(4.4.6) |
  |------------|------:|---------------------:|----------------------:|------------:|------:|
  | Process    |    15 |              718 ns |              2173 ns |     1180 ns | 2522 ns |
  | Thread     |    10 |              251 ns |               666 ns |      458 ns | 1055 ns |
  | RPC        |     9 |              275 ns |               642 ns |      491 ns |  771 ns |
  | File       |     7 |              228 ns |               402 ns |      385 ns |  872 ns |
  | Registry   |     6 |              207 ns |               300 ns |      324 ns |  467 ns |

~~Spoiler - if you want to (near) guarantee best performance you can let the caller provide the expected index rather than guessing.  (My library has this option).~~

Added expected index hint option.

### DIFF
--- a/O365.Security.Native.ETW.Debug.nuspec
+++ b/O365.Security.Native.ETW.Debug.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>Microsoft.O365.Security.Native.ETW.Debug</id>
-        <version>4.4.8</version>
+        <version>4.4.9</version>
         <title>Microsoft.O365.Security.Native.ETW Debug - managed wrappers for krabsetw</title>
         <authors>Microsoft</authors>
         <owners>Microsoft</owners>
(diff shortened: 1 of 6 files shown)


kylereedmsft:
Thanks for pulling in the hint changes. Are the numbers in the table for the current PR (with hints)? Just so I understand, worst-case is reverse order without hints, right?

kylereedmsft (approved):

jdu2600:
The above table is the hint-free case only.

I provided some addition comparisons in https://github.com/microsoft/krabsetw/pull/273#discussion_r3062265173.

The relevant portions for this PR were -

Accessing every third property -
 | Event | Props | last+1 | last+1 reverse | hinted (+reverse ) |
  |---|:-:|---:|---:|---:|
  | Process | 6 of 16 | 482.9 | 1105.4 | 498.3 |
  | Thread | 4 of 10 | 199.9 | 309.6 | 183.7 |
  | RPC | 3 of 9 | 166.7 | 220.3 | 155.1 |
  | File | 3 of 7 | 208.3 | 237.4 | 197.5 |
  | Registry | 2 of 6 | 111.3 | 120.4 | 109.4 |

Accessing the middle property only -
| Event type |  last+1  | hinted |
  |---|---|---|
  | Process |  184.4 | 145.2 |
  | Thread | 130.7 | 127.0 |
  | File | 97.6 | 93.0 |
  | Registry | 98.8 | 93.3 |
  | RPC | 139.5 | 127.5 |

PR #276, opened 2026-04-06
Rework parser: eager cache, hinted scan, wstring_view

kylereedmsft:
- Replace lazy deque cache with vector<property_info> fully populated on first access (ensure_cache_populated)
- Add hinted linear scan for name->index resolution; sequential access is O(1) per call, callers can also provide explicit hints
- Add parse/try_parse/view_of overloads accepting a ULONG hint parameter
- Convert all property name parameters from const std::wstring& to std::wstring_view to avoid temporary allocations
- Chain no-hint overloads to hint versions to eliminate code duplication

### DIFF
--- a/krabs/krabs/parser.hpp
+++ b/krabs/krabs/parser.hpp
@@ -80,6 +80,15 @@ namespace krabs {
 
         /**
          * <summary>
+         * Attempts to retrieve the given property by name and type,
+         * starting the name scan at the given hint index.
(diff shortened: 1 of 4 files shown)


kylereedmsft:
We decided that for our normal use cases (sparse access), eager caching adds more overhead even though it's better for full record access. We're going with #275 instead which includes the hinting from this PR.

PR #281, opened 2026-09-13
Add error handling callback for enabling user-trace providers

Serious-senpai:
**Current issue:** When a user trace has multiple providers enabled, an exception thrown by [`EnableTraceEx2`](https://learn.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-enabletraceex2) for any one of the providers causes the entire trace to fail when `trace.start()` is called.

This behavior isn't always desirable. For example, a monitoring product may want its trace to continue receiving ETW events on a best-effort basis, even if some providers cannot be subscribed to.

**This PR solution:** Provide an exception callback to handle errors returned by [`EnableTraceEx2`](https://learn.microsoft.com/en-us/windows/win32/api/evntrace/nf-evntrace-enabletraceex2) during `trace.start()`. The callback is implemented by library users, allowing them to decide how to handle the failure. They may choose to abort the trace entirely, or continue starting the trace while acknowledging that some providers could not be subscribed to.


Serious-senpai:

swannman (approved):

